PR #7, opened 2020-08-03
Add colors to icons in buffer line using lua web dev icons

Akin909:
tabline-colors

:thinking: not sure how I feel about these colors :laughing: 

Update (much nicer):
colors-nvim-devicons


Shatur95:
There is a [great plugin](https://github.com/lambdalisue/glyph-palette.vim) for it, maybe is it better to provide integration with it?

Akin909:
Thanks for the heads up @Shatur95, it looks cool. Although I want to stick with lua dependencies for speed where possible, that one is written in vimscript (not that I'm benchmarking stuff). I'm already using https://github.com/kyazdani42/nvim-web-devicons for this since I can require it directly in lua

Shatur95:
I understand, I prefer lua too. But this `glyph-palette.vim` is only global highlighter for nerd font glyphs, it not assigns icons itself. If i understood correctly,  https://github.com/kyazdani42/nvim-web-devicons only for icons.
Maybe it would be better to use a separate lua plugin that just highlight all glyphs for specified buffer (as `glyph-palette.vim`)?

Akin909:
I'm already making a PR to `nvim-web-devicons` to add highlights I talked with the author about adding that a while ago so I can use it here. The plan is for that package to add highlights as well as icons

Issue #9, opened 2020-09-06
show bufferline only when two or more buffers are open


astier:
Hi. First of all thank you for the great plugin. Out of all bufferline-plugins I finally found one which does a good job. I especially like the mouse support. Would it be possible to show the bufferline only when two or more buffers are open?

astier:
So I figured out that bufferline depends in a weird way on the `showtabline` setting. I partially achieved my goal with the following autocmd:

```vim
autocmd group vimenter,bufadd * if len(getbufinfo({'buflisted':1})) > 1 | set showtabline=2
autocmd group bufdelete * if len(getbufinfo({'buflisted':1})) < 2 | set showtabline=1
```

The first autocmd works. It I open nvim with only one buffer the bufferline is not shown. It I then open a second buffer the bufferline is shown.

The second autocmd does not work for some reason. If I close the second buffer then the bufferline is still shown. If I then set manually `set showtabline=1` it isnt shown anymore.

astier:
### Minimal vimrc

```vim
set runtimepath+=~/.local/share/nvim/plugins/nvim-bufferline.lua/
autocmd bufadd,vimenter * if len(getbufinfo({'buflisted':1})) > 1 | set showtabline=2
autocmd bufdelete * if len(getbufinfo({'buflisted':1})) < 2 | set showtabline=1
lua require'bufferline'.setup()
set showtabline=1
```

For some reason it doesn't work if I put `set showtabline=1` before `lua require'bufferline'.setup()`

akinsho:
@astier thanks for trying out the plugin 👍🏿 .
I use the `showtabline` function primarily because it's the only way in vim to specify tabline visibility even though it was originally intended for use with *tabs* rather than buffers so the `showtabline` setting doesn't really apply in the same way. `2` makes it show all the time and `1` makes it show if there are 2 or more `tab pages` rather than buffers specifically.

To add something like that it would need to have specific logic. Tbh I personally don't think I'd ever use such an option but if you'd like to have a go at implementing it I'd be fine with that, if it didn't add too much complexity.

I think re implementation you'd need to pass in another option in the user config passed to the `setup` and then register some autocommands to watch `BufEnter` (or something) and check the bufferlist and if it's ever less than 2 then hide the line by setting `showtabline=0`

astier:
[...]

PR #12, opened 2020-10-01
Add option to disable tab when only one buffer is opened

ray-x:


lua/bufferline.lua
```diff
@@ -543,6 +543,16 @@ local function bufferline(preferences)
   local buffers = {}
   local tabs = get_tabs()
   preferences.options.view = current_mode
+  if preferences.options.always_show_tab == false then
+    numItem = 0
```
akinsho, line 547: @ray-x I *think* what you are trying to do here is get the number of open buffers, in which case you should be able to use `table.getn(buf_nums)` rather than looping through and incrementing a count

lua/bufferline.lua
```diff
@@ -594,6 +604,7 @@ local function get_defaults()
       mappings = false,
       show_buffer_close_icons = true,
       enforce_regular_tabs = false,
+      always_show_tab = true,
```
akinsho, line 607: I would call this `always_show_bufferline` since I wasn't sure what it was meant to do since it was called `always_show_tab` I though you were trying to hide tab indicators initially

lua/bufferline.lua
```diff
@@ -686,6 +697,11 @@ function M.setup(prefs)
     {"VimEnter", "*", [[lua __setup_bufferline_colors()]]};
     {"ColorScheme", "*", [[lua __setup_bufferline_colors()]]};
   }
+  if preferences.options.always_show_tab == false then
```
akinsho, line 700: I think the more lua way to right this is
```lua
if not preferences.options.always_show_tab then
```

lua/bufferline.lua
```diff
@@ -686,6 +697,11 @@ function M.setup(prefs)
...
+    -- toggle tabline
+    table.insert(autocommands, {"VimEnter,BufAdd,TabEnter", "*", "if len(getbufinfo({'buflisted':1}))>1 | set showtabline=2 | else | set showtabline=0 | endif"})
```
akinsho, line 702: I haven't tested this will do so later today, but someone else raised this in #9 and it seemed that `getbufinfo` wasn't working on `BufDelete` I'm assuming it works correctly on `TabEnter` and `BufAdd`?

nit: can you please indent this line since it seems to be really long 🙏🏿

akinsho:
@ray-x thanks for the PR 👍🏿 I appreciate you taking the time to add this 🚀 . I've left a few style comments, lemme know if you have any questions or comments

lua/bufferline.lua
ray-x: thanks, updated code to use table.getn(buf_nums)

lua/bufferline.lua
ray-x: yes, `always_show_bufferline` is better naming. code updated.

lua/bufferline.lua
ray-x: Thanks. Code updated

lua/bufferline.lua
ray-x: when buffer deleted, TabEnter will also be trigger, so `len(getbufinfo({'buflisted':1}))>1` should still work. I did run some test and it behave correct.
So maybe still use this ?
```vim
    table.insert(autocommands, {"VimEnter,BufAdd,TabEnter", "*",
      "if len(getbufinfo({'buflisted':1}))>1 | set showtabline=2 | else | set showtabline=0 | endif"})
```

ray-x:
@akinsho Thanks for the feedbacks. Code updated based on your feedbacks.

akinsho:
Tried this out locally and seems to work well. Thanks for the contribution 👍🏿

PR #15, opened 2020-10-16
add buffer_close_icon support

glepnir:
@akinsho  cc


akinsho:
@glepnir thanks for the PR. I'm actually trying to migrate away from the variables configured via vimscript and have all the user config live in one lua table passed in on setup.

I already have an example of how to pass in a custom icon [here](https://github.com/akinsho/nvim-bufferline.lua/blob/master/lua/bufferline.lua#L604). Could you change this to do a similar thing. So add a new `buffer_close_icon` option to the user config object passed to startup and use that value.

At some point when I have time I'm going to remove all the `g:bufferline_*` variables

glepnir:
@akinsho  easy to do that.

glepnir:
@akinsho  Please review . the new version that remove all the `helpers.get_plugin_variable` function. we use table to store the  icon value.

glepnir:
Test success on my local  @akinsho can merge?

akinsho:
@glepnir I'm at work atm I will merge it when later on today 👍🏾

akinsho:
Awesome just tried it out during my lunch break and it looks good 👍🏿 thanks for fixing all the other usages of `g:variables`

glepnir:
No thanks. My pleasure.

PR #21, opened 2020-10-26
Feature/custom separator

elianiva:
I think it would be nice if we can define our own separator so I added an option to do that by passing a table to `separator_style`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,9 @@ require'bufferline'.setup{
     max_name_length = 18,
     tab_size = 18,
     show_buffer_close_icons = true | false,
-    separator_style = "thick" | "thin",
+    -- can also be a table containing 2 custom separator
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -139,7 +139,9 @@ require'bufferline'.setup{
     max_name_length = 18,
     tab_size = 18,
     show_buffer_close_icons = true | false,
-    separator_style = "thick" | "thin",
+    -- can also be a table containing 2 custom separator
```
akinsho, line 142: `separator` -> `separators`

elianiva: fixed

akinsho:
@elianiva thanks for the PR will merge this now, although I don't really plan on adding many more configuration options FYI, I think things are getting kind of unwieldy with the level of customisation available already.

elianiva:
Ah, I see. Thanks for making this awesome plugin!

PR #22, opened 2020-10-28
fix: sorting by directory

romgrk:
Typo there

### DIFF
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -106,7 +106,7 @@ end
 --- @param buf_a Buffer
 --- @param buf_b Buffer
 local function sort_by_directory(buf_a, buf_b)
-  return fnamemodify(buf_a.path, ':h') < fnamemodify(buf_b.filename, ':h')
+  return fnamemodify(buf_a.filename, ':h') < fnamemodify(buf_b.filename, ':h')
 end
 
 --- sorts a list of buffers in place


lua/bufferline.lua
```diff
@@ -106,7 +106,7 @@ end
 --- @param buf_a Buffer
 --- @param buf_b Buffer
 local function sort_by_directory(buf_a, buf_b)
-  return fnamemodify(buf_a.path, ':h') < fnamemodify(buf_b.filename, ':h')
+  return fnamemodify(buf_a.filename, ':h') < fnamemodify(buf_b.filename, ':h')
```
akinsho, line 109: Thanks for the PR, actually they should both be `path` which is the full file path whereas filename is just the filename

romgrk: Right, actually I've just left `buf.path` for both, because otherwise if you `fnamemodify('file-in-cwd.vim', ':h')` you end up with `'.'`. Also even if both files are in the same directory, it's probably better to have them sorted alphabetically by their filename.

romgrk: While I'm at it, I've added another commit to sort file with paths relative to cwd first, the other ones last.

akinsho: This looks good thanks for looking into it, I'm away from my machine atm but will look at it properly later

akinsho:
@romgrk I tried this out locally and made a few changes based on some issues, the final result looks like
```lua
-- This needs to be defined before usage
local function is_relative_path(path)
  return fnamemodify(path, ':p') ~= path
end

--- @param buf_a Buffer
--- @param buf_b Buffer
local function sort_by_directory(buf_a, buf_b)
  local ra = is_relative_path(buf_a.path) -- <- this should reference the path
  local rb = is_relative_path(buf_b.path)
  if ra and not rb then
    return false <-- we need to return booleans here
  elseif rb and not ra then
    return true
  else
    return buf_a.path < buf_b.path
  end
end
```

lua/bufferline.lua
```diff
@@ -106,9 +106,22 @@ end
 --- @param buf_a Buffer
 --- @param buf_b Buffer
 local function sort_by_directory(buf_a, buf_b)
-  return fnamemodify(buf_a.path, ':h') < fnamemodify(buf_b.filename, ':h')
+  local ra = is_relative_path(buf_a)
+  local rb = is_relative_path(buf_b)
+  if ra and not rb then
+    return -1
```
akinsho, line 112: These need to be booleans -> `false`

lua/bufferline.lua
```diff
@@ -106,9 +106,22 @@ end
...
+  end
+  if rb and not ra then
+    return +1
```
akinsho, line 115: same here -> `true`

lua/bufferline.lua
```diff
@@ -106,9 +106,22 @@ end
 --- @param buf_a Buffer
 --- @param buf_b Buffer
 local function sort_by_directory(buf_a, buf_b)
-  return fnamemodify(buf_a.path, ':h') < fnamemodify(buf_b.filename, ':h')
+  local ra = is_relative_path(buf_a)
```
akinsho, line 109: This is passing in the whole buffer not just the path. Should be `buf_a.path`

lua/bufferline.lua
```diff
@@ -106,9 +106,22 @@ end
...
+  if ra and not rb then
+    return -1
+  end
+  if rb and not ra then
+    return +1
+  end
+  return buf_a.path < buf_b.path
 end
 
+local function is_relative_path(path)
```
akinsho, line 120: This should be defined before it's used 🙏🏿 (at least the lsp warns me it should and I prefer it for sanity's sake 😄 )

akinsho (changes requested):
Thanks again for your work on this, just a few issues before merging (currently this errors/crashes the plugin)

romgrk:
Fixed, sorry I didn't had enough time to test, I just wanted to fix the typo but sometimes I can't resist when I see something that could be better ^^

Btw I like your extension very much, I've integrating some elements of it in barbar.nvim, the padding and closing icons, it looks much nicer now :)

Screenshot from 2020-10-29 20-04-54

One thing I added is auto-sizing tab width, the tabs use less padding when there is not enough space for all tabs. Could be nice to implement in nvim-bufferline.lua as well.

PR #33, opened 2020-12-03
fix nil prefs error

ellisonleao:
Recent changes is breaking the plugin for me, the i realized that part was not handling a possible nil prefs obj. That fixed the issue on my side, not sure if its also just something on my side or an actual bug.

My configs are not using any custom prefs in `require('bufferline').setup()` .


akinsho:
Thanks for the PR @ellisonleao 🙌🏿 didn't notice that bug

PR #38, opened 2021-01-07
Add functionality to filter buffers

cooper-anderson:
This PR adds functionality to filter what buffers are visible using a function that the user can specify in their options. 

This function runs for each buffer that is open. It currently takes in the buffer's ID as a parameter, but looking at the sort function, perhaps a reference to the buffer itself could be better. The user is able to specify a return value of `true`, `false` or `nil` to set the visibility (`nil` uses the default visibility).

Here is an example of how this feature can be used:
image
Here is the same instance vim instance after pressing `gt`
image

Currently this feature does not apply when the user has their `view` option set to `multiwindow`.

TODO:

- [x] Add basic filtering method
- [x] Allow the user to override the default method in their options
- [x] Add documentation for filtering buffers to the README
- [x] Add a gif of the feature being used

TO-POSSIBLY-DO

- [ ] Redo the example gif so that it uses the same esthetics as the existing gifs
- [ ] Add ability to have multiple filtering methods
	- [ ] Add separate function used to determine which filtering method to use

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,6 +169,9 @@ require'bufferline'.setup{
     sort_by = 'extension' | 'relative_directory' | 'directory' | function(buffer_a, buffer_b)
       -- add custom logic
       return buffer_a.modified > buffer_b.modified
+    end,
+    filter = function(buffer_id)
+      -- add custom logic, or return `nil` to use the default
     end
   }
 }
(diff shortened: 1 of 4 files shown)


akinsho:
@cooper-anderson thanks for working on this 👍🏿 I really like the approach (versatility of a using a function) and the examples + docs 🚀 . I'll try it out later this evening.

cooper-anderson:
@akinsho Awesome! Glad you like it! I've known about this plugin for a decent amount of time but instead of just using it I thought to write my own... in vimscript... I recently moved all my neovim configs to lua so I decided to give this plugin a try, and I have been really enjoying it!

akinsho:
@cooper-anderson I've just had a chance to try it out and I think there are a lot of potential use cases. One issue I encountered is that I use nvim almost exclusively via sessions and so currently since the filter prevents the initial population of the bufferline buffer list, if you open a session with the filter set to something like
```lua
      filter = function(bufnr)
        return vim.bo[bufnr].filetype ~= "lua"
      end
```
say you have 10 buffers open and half are lua, when I start the session if I start on a non lua filetype my bufferline doesn't show all the non lua files immediately. As I navigate them bufferline "discovers" them then adds them in memory.

also because we are no longer filtering properly say you load a session with a filter in your config all the previously unloaded buffers show up.

Here's an example of a bunch of zombie buffers not being filtered out from a session
image

I think the filtering should always happen *after* the initial filter which is necessary for removing invalid/unloaded etc. buffers

I managed to solve the first problem by moving the filter to always happen afterwards. The first issue is weird since once that is moved,the filter still behaves incorrectly when you start vim and there are already files open, you might be able to reproduce this running `nvim <list-of-mixed-filetypes>`

akinsho:
sorry that wasn't the best possible description, it's best if you have a go by creating a session manually populating it then restoring it with some basic filter to remove a filetype or something and see the results. I'm not 100% sure what's going on and haven't dug too deeply into it.

cooper-anderson:
Interesting, in each of my test scenarios, I had the filtering set to only be on if the current tab had a `t:` variable set. If that variable was not set, the filter returned `nil` so it used the normal filtering method. It looks like in your example it only returns `true` or `false`, so all you will ever see are the lua filetypes. Having the default case return `nil` would also fix that issue of showing every single buffer, even the ones you could not see normally.

For fixing the initial load with files already open, perhaps there could be an autocmd that just refreshes the first time vim is loaded. I'll have to experiment with that.

Thanks for the feedback!

akinsho:
There's a `SessionLoadPost` autocommand for when a session is loaded but tbh that's just one scenario and it's interesting/weird that the initial population of the list doesn't include the all the non lua files right from the start 🤔

cooper-anderson:
Yeah you think that if it is happening it would happen if there wasn't a filter too... I'll try messing around with it for a bit and see what more we can figure out.

cooper-anderson:
I'm gonna close this for now, as it is just sitting here. I might try to come up with a better idea on how to implement it without running into the same issues in the future.

PR #39, opened 2021-01-13
Add LSP diagnostics indicator

akinsho:
lsp_error

This PR uses the recently added `vim.lsp.diagnostic.get_all()` to get diagnostics for attached lsp buffers and display them. Ideally this should also highlight differently if the errors are all warning level.

### Todo
- [x] Make sure this is fully gated and doesn't break for anyone who isn't opted in

### Follow up
* Add `coc` support if/when I figure out how to get all diagnostics
* add warning highlights


cooper-anderson:
I did something similar manually, I think it was visible in my pictures uploaded for #38. I do like how simple this way of displaying is though. I will try it out later, but does it show the status of buffers that are not visible? Or does it unload LSP information when a buffer is hidden.

akinsho:
That's cool @cooper-anderson just checked out your screenshot 👍🏾. Tbh this is almost certainly not the final UI for it. Just dumped it in there so I could have a working version to see how it behaves and get inspired.

I'm thinking maybe a red underline on the buffer, we'll see. 

Re the behaviour of the function it seems to return a map for all correctly loaded buffers with an attached last client.

PR #45, opened 2021-01-26
Fix rendering problem of file icons in Windows Terminal

LimbicSys:
In Windows Terminal, file's icon shows only left part when the buffer is selected. I print the string of the tabline,  finding that the icon's right half is override by other characters.  (The icon I use is two characters width.)  So I make the following changes:
  1. Move the white space right before the filename to the position right
     after the icon.
  2. Insert a white space between the letter when BufferLinePick called.

I use it in both Windows Terminal and vscode's terminal and both work properly.

### DIFF
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -282,8 +282,8 @@ local function add_prefix(context)
   local length = context.length
 
   if state.is_picking and buffer.letter then
-    component = hl.pick .. buffer.letter .. hl.background .. component
-    length = length + strwidth(buffer.letter)
+    component = hl.pick .. buffer.letter .. padding .. hl.background .. component
+    length = length + strwidth(buffer.letter) + strwidth(padding)
   elseif buffer.icon then
     local icon_highlight = highlight_icon(buffer, hl.buffer)
     component = icon_highlight .. hl.background .. component
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -210,7 +210,7 @@ local function highlight_icon(buffer, background)
     local guifg = colors.get_hex(hl, "fg")
     highlights.set_one(new_hl, {guibg = background.guibg, guifg = guifg})
   end
-  return "%#" .. new_hl .. "#" .. icon .. "%*"
+  return "%#" .. new_hl .. "#" .. icon .. " %*"
 end
 
 --- "▍" "░"
@@ -280,7 +280,6 @@ local function add_prefix(context)
   local buffer = context.buffer
   local hl = context.current_highlights
   local length = context.length
-
   if state.is_picking and buffer.letter then
     component = hl.pick .. buffer.letter .. padding .. hl.background .. component
     length = length + strwidth(buffer.letter) + strwidth(padding)
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -210,7 +210,7 @@ local function highlight_icon(buffer, background)
     local guifg = colors.get_hex(hl, "fg")
     highlights.set_one(new_hl, {guibg = background.guibg, guifg = guifg})
   end
-  return "%#" .. new_hl .. "#" .. icon .. " %*"
+  return "%#" .. new_hl .. "#" .. icon .. padding .. "%*"
 end
 
 --- "▍" "░"
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -280,13 +280,14 @@ local function add_prefix(context)
   local buffer = context.buffer
   local hl = context.current_highlights
   local length = context.length
+
   if state.is_picking and buffer.letter then
     component = hl.pick .. buffer.letter .. padding .. hl.background .. component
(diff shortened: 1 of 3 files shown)


lua/bufferline/buffers.lua
```diff
@@ -81,9 +81,11 @@ function M.Buffer:new(buf)
     if lua_devicons_loaded then
       buf.icon, buf.icon_highlight =
         webdev_icons.get_icon(buf.path, buf.extension, {default = true})
+      buf.icon = buf.icon .. ' '
```
akinsho, line 84: I don't think the highlighting should go here, I don't really want UI specific things inside of this file having the icon here at all is something I'm probably going to refactor since this should just be a model file. This change should go to where the icon is added when creating the buffer string i.e. `add_prefix` function in `bufferline.lua` not here

lua/bufferline/duplicates.lua
```diff
@@ -112,11 +112,8 @@ function M.component(context)
         return truncate(dir, depth, options.max_prefix_length)
       end
     )
-    component = padding .. hl.duplicate .. dir .. hl.background .. component
-    length = length + strwidth(padding .. dir)
-  else
-    component = padding .. component
-    length = length + strwidth(padding)
+    component = hl.duplicate .. dir .. hl.background .. component
+    length = length + strwidth(dir)
```
akinsho, line 116: Was this change to remove the extra space that was being added before so there aren't two spaces?

lua/bufferline/buffers.lua
LimbicSys: Yes, your are right. I will  move this to `add_prefix`.

lua/bufferline/duplicates.lua
LimbicSys: Yes, otherwise there will be two spaces. And it looks a liitle strange.

akinsho:
Tbh I was a little concerned this would break things for other users but it seems to work the same as far as I can tell. If you just move the actual appending of the space to the icon where the icon is added to the string then this will be ok 👍🏿

lua/bufferline.lua
```diff
@@ -210,7 +210,7 @@ local function highlight_icon(buffer, background)
     local guifg = colors.get_hex(hl, "fg")
     highlights.set_one(new_hl, {guibg = background.guibg, guifg = guifg})
   end
-  return "%#" .. new_hl .. "#" .. icon .. "%*"
+  return "%#" .. new_hl .. "#" .. icon .. " %*"
```
akinsho, line 213: @LimbicSys can you please add this after the `%*`. As part of it that is not very clear. The best place for is after the `icon_highlight` using the `padding` variable rather than just a raw space character. Or if will that not work on windows then you can add it as `icon .. padding .. "%*"`

LimbicSys: `"%*" .. padding` will cause the similar problem. Now the selected buffer shows correctly, but unselected ones' icons show only haft part. So I just replce the space with `padding`.

lua/bufferline.lua
```diff
@@ -280,7 +280,6 @@ local function add_prefix(context)
   local buffer = context.buffer
   local hl = context.current_highlights
   local length = context.length
-
```
akinsho, line 283: Can you leave this new line in place 🙏🏿

lua/bufferline.lua
```diff
@@ -280,10 +280,9 @@ local function add_prefix(context)
...
   if state.is_picking and buffer.letter then
-    component = hl.pick .. buffer.letter .. hl.background .. component
-    length = length + strwidth(buffer.letter)
+    component = hl.pick .. buffer.letter .. padding .. hl.background .. component
+    length = length + strwidth(buffer.letter) + strwidth(padding)
   elseif buffer.icon then
     local icon_highlight = highlight_icon(buffer, hl.buffer)
     component = icon_highlight .. hl.background .. component
```
akinsho, line 288: You will also need to update the `length` of the `icon` below since this now includes `padding` ie. `strwidth(buffer.icon .. padding)`

akinsho:
@LimbicSys thanks for the tweaks, once the last changes are I suggested are done I'll merge this.

I think it's important to note that I don't want to get into the habit of restructuring stuff obscurely because of certain terminal emulators so I think it's worth raising this on the window terminal repo so it can be fixed at the source. If it is ever necessary to undo this because it makes future changes more complicated I might have to, so will be better if they fix the issue.

lua/bufferline.lua
LimbicSys: Sorry, I deleted it by mistake.

LimbicSys:
@akinsho I do want to raise the issue to Windows Terminal, but I cannot reproduce this problem in a simple way. And it seems that it is related to how neovim highlights  and renders characters, because the icon shows correctly when I use vim-devicons without highlighting.

akinsho:
Ok LGTM, will merge this now. Thanks for your work on this @LimbicSys

PR #48, opened 2021-01-31
Escape percent symbol in filenames

### LINKED ISSUE #47: Special characters in file name are not escaped properly.
When the file name contains special characters, the buffer line will not present the proper name. Such as `a%.txt` is shown as `aa%.txtxt`. This issue is quite subtle and rare(only when special characters exist), but I would be grateful if it got fixed.

akinsho:
Should fix #47. @AnthonyK213 can you try this and see if it fixes your issue?


AnthonyK213:
Yes, it works well. Thanks!

PR #52, opened 2021-02-22
fix nil prefs objects

ellisonleao:


akinsho:
@ellisonleao thanks completely missed that edge case 🙏🏿

akinsho:
@ellisonleao I just tweak the conditional here slightly so it's combined with the statement below on master, can you see if this still works for you (it should tested it out but just want to make sure)

ellisonleao:
@akinsho working just fine, thanks again for the quick fix!

Issue #54, opened 2021-03-05
Add command to sort buffers


Charliiee:
## Add possibility to sort buffers on the fly

Expose a module function to sort buffers. I think even creating some commands for it would be a good idea too, as well as keymapping examples.

E.g.:

![image](https://user-images.githubusercontent.com/13454744/110108106-88434180-7d8a-11eb-89e9-f37295f7d1ef.png)

After sorting by directory:

![image](https://user-images.githubusercontent.com/13454744/110108213-a5781000-7d8a-11eb-8cc2-572388b264eb.png)

akinsho:
@Charliiee thanks for checking out the plugin and raising this issue. I noticed some of the alternative plugins provide similar functionality, I'm not against the idea and would happily accept a  PR for anyone looking to implement this, but I've spread myself pretty thin across quite few projects so I'm not likely too do this one myself for a while, I just won't have the time.

Happy to also provide some advice if anyone want's to look into implementing this, most of the logic for sorting is already done and it will mainly be about triggering a particular sort during runtime.

Charliiee:
Great! I'm already working on it then! :+1: 

I'll just test the goto buffer feature after mappings to see if things work properly and then provide some examples in docs before submitting a PR

By the way, I don't know if it's only me but the plugin `docs` is not appearing for me on nvim `:h`

akinsho:
Awesome @Charliiee thanks a lot, looking forward to it 🚀 . Regarding the help files they should be generated by your package manager automatically  based on the contents of the doc file. Which package manager do you use?

Charliiee:
I'm using vim-plug

akinsho:
@Charliiee what are you typing into the `:h`? the main doc is under `:h nvim-bufferline.lua` and the headings are of the format `:h bufferline-styling` or `:h bufferline-<subject>` you can also try `:helpgrep bufferline` to see if anything show up. I use it with `packer.nvim` and it seems to build just fine on my personal dev machine as well as my work machine.

Charliiee:
OK, I ran a full update of plugins (including vim-plug) and the helptag seems to be working fine now, thanks

PR #55, opened 2021-03-08
Sort buffers

Charliiee:
Expose function to sort buffers

Solves #54


akinsho:
@Charliiee this looks good 👍🏿 thanks for adding it 💯 . One thing I think it would be nice to add is a command for the sorting, so users using vimscript or who just prefer it can actually type it out in the commandline rather than trying to write out a lua function call which can be a pain
something like

`BufferLineSortByDirectory` and `BufferLineSortByExtension` there's some prior art for the bufferline cycle commands `vim.cmd('command! BufferLineMoveNext lua require"bufferline".move(1)')`

I also think this is worth adding to the README somewhere since I think other people would probably be interested in this

akinsho:
Thanks for the contribution @Charliiee 🚀

PR #62, opened 2021-03-16
Add option to omit close icon

samyak-jain:
Added the ability to make the close button at the far right optional.

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,6 +158,7 @@ require'bufferline'.setup{
       return "("..count..")"
     end
     show_buffer_close_icons = true | false,
+    show_close_icon = true | false,
     show_tab_indicators = true | false,
     persist_buffer_sort = true, -- whether or not custom sorted buffers should persist
     -- can also be a table containing 2 custom separators
--- a/doc/bufferline-lua.txt
+++ b/doc/bufferline-lua.txt
@@ -64,6 +64,7 @@ The available settings are: >
             return "("..count..")"
             end
             show_buffer_close_icons = true | false,
+	    show_close_icon = true | false,
             show_tab_indicators = true | false
             persist_buffer_sort = true, -- whether or not custom sorted buffers should persist
             -- can also be a table containing 2 custom separators
--- a/doc/bufferline-lua.txt
+++ b/doc/bufferline-lua.txt
@@ -64,7 +64,7 @@ The available settings are: >
             return "("..count..")"
             end
             show_buffer_close_icons = true | false,
-	    show_close_icon = true | false,
+            show_close_icon = true | false,
             show_tab_indicators = true | false
             persist_buffer_sort = true, -- whether or not custom sorted buffers should persist
             -- can also be a table containing 2 custom separators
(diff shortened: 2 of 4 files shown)


lua/bufferline/config.lua
```diff
@@ -45,6 +45,7 @@ function M.get_defaults()
       max_name_length = 18,
       mappings = false,
       show_buffer_close_icons = true,
+			show_close_icon = true,
```
akinsho, line 48: Can you switch the indentation you are using to spaces instead of tabs, the project uses 2 spaces to indent 🙏🏿 . This should also line up with the option above

lua/bufferline.lua
```diff
@@ -587,7 +587,11 @@ local function render(bufs, tbs, prefs)
     line = join(line, hl.background.hl, icon)
   end
 
-  return join(line, hl.fill.hl, right_align, tab_components, hl.tab_close.hl, close)
+	if options.show_close_icon then
```
akinsho, line 590: rather than change it like this I think it would be better to make the change at line 543, something like
```lua
local close, close_length = "", 0 -- initialise as zero values
if options.show_close_icon then
  close, close_length = render_close(options.close_icon)
end
```
That way close will be empty and that line doesn't have to change items in the list

doc/bufferline-lua.txt
```diff
@@ -64,6 +64,7 @@ The available settings are: >
             return "("..count..")"
             end
             show_buffer_close_icons = true | false,
+	    show_close_icon = true | false,
```
akinsho, line 67: Similar to the other command can you use spaces here and line up the indentation with the line above 🙏🏿

README.md
```diff
@@ -158,6 +158,7 @@ require'bufferline'.setup{
       return "("..count..")"
     end
     show_buffer_close_icons = true | false,
+    show_close_icon = true | false,
```
akinsho, line 161: Thanks for updating the docs 🚀

akinsho (changes requested):
Thanks for raising this PR to fix the issue @samyak-jain, the help is much appreciated. I've left some comments regarding indentation and a note on the implementation.

lua/bufferline/config.lua
samyak-jain: Done! Looks I hadn't configured neovim identation properly.

samyak-jain:
I've made the changes. Let me know if this works.

lua/bufferline.lua
```diff
@@ -540,7 +540,10 @@ local function render(bufs, tbs, prefs)
   local hl = prefs.highlights
   local right_align = "%="
   local tab_components = ""
-  local close, close_length = render_close(options.close_icon)
+	local close, close_length = "", 0
+	if options.show_close_icon then
+		close, close_length = render_close(options.close_icon)
+	end
```
akinsho, line 546: @samyak-jain thanks for fixing the other indentation 🙏🏿 looking at this block it still seems to be indented more than the lines above it. The same goes for the return at the bottom of the function. Once those are done I'll be happy to merge this. Thanks again 👍🏿

samyak-jain: Yup. Sorry about that. Should be fine now.

akinsho (approved):
Awesome this looks great now thanks for the fixes 👍🏿  will merge this now

PR #63, opened 2021-03-17
Allow users to specify a custom filter function

akinsho:
This PR will provide the ability for a user to specify a custom filter function to determine which buffers should not be shown in the bufferline. A user could always do something like `set bufnolisted` etc for a filetype to prevent it from showing up but with a custom filter function they can now do it based on more complex set of circumstances/data.

This is quite similar/relates to #38, cc @cooper-anderson if you are still interested

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,22 @@ require'bufferline'.setup{
     diagnostics_indicator = function(count, level)
       return "("..count..")"
     end
+    -- NOTE: this will be called a lot so don't do any heavy processing here
+    custom_filter = function(buf_number)
+      -- filter out filetypes you don't want to see
+      if vim.bo[buf_number].filetype ~= "<i-dont-want-to-see-this>" then
+        return true
+      end
+      -- filter out by buffer name
+      if vim.fn.bufname(buf_number) ~= "<buffer-name-I-dont-want>" then
+        return true
+      end
+      -- filter out based on arbitrary rules
+      -- e.g. filter out vim wiki buffer from tabline in your work repo
+      if vim.fn.getcwd() == "<work-repo>" and vim.bo[buf_number].filetype ~= "wiki" then
+        return true
+      end
+    end,
     show_buffer_close_icons = true | false,
     show_close_icon = true | false,
     show_tab_indicators = true | false,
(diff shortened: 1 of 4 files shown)


cooper-anderson:
Definitely interested, though I am a little pressed for time currently. I will definitely look into contributing if I find the time.

akinsho:
@cooper-anderson that's alright I'm fine to do the implementation, granted it basically already works. The main thing I thought I'd get your input on was the tri-state thing you had in your PR i.e. `nil`, `true` and `false` all meaning something in the filtering. I'm not sure I understood what that was for in the end? My thinking is a buffer is filtered or not, what's the other state that nil would represent. You mentioned default visibility in your PR but I'm not sure what that means?

cooper-anderson:
Ah, okay I will try to explain what I meant by "default visibility" a bit more clearly, but I now think that using a simple `true`/`false` system would be more efficient.

Instead of just iterating through the `buflisted` buffers, my PR iterated through all buffers (more on this later). The way nvim-bufferline is now can be thought of as using a filter that always returns `nil`. This causes it to use whatever vim would classify as a "listed buffer".

However, if the user wanted to have, for example, tabs with different listed (possibly non-visible) buffers, they can create a filter function that checks for if that given tab has some tab-variable, and then return `true` or `false` depending on if they want a particular buffer listed, as opposed to letting vim choose by way of `nil`.

I have since found out that often times there are quite a few buffers behind the scenes that are `nobuflist`ed on purpose, and having to loop through all of those extra buffers would just add unneeded complexity. To get around this, perhaps nvim-bufferline still should loop only on the `buflist`ed buffers, but if the user wants some buffer listed that is `nobuflist` by default (like help pages) they could just write an autocmd that sets that option.

So yes, a simple `true`/`false` would probably be best, and the default filter function could just return `true`.

Let me know if there is anything that is still unclear with my explanation :+1:

akinsho:
Ah thanks for clarifying 👍🏿 yeah think starting with `buflisted` + `true` and `false` is simplest to start and if a user wants something more custom re what is and isn't listed they can use autocommands to achieve that without having to do a bunch of extra filtering in bufferline

PR #67, opened 2021-03-25
Return diagnostics table with each level count (errors, warnings and info) for the diagnostics_indicator

arsaniwilliam:
Return a dictionary of error level and number of errors for each level.
It's backwards compatible and should allow for a custom diagnostic count such as this:
 ```lua
 config = {
  ....
  diagnostics_indicator = function(_, _, x)
      local s = " "
      for e, n in pairs(x) do
        local sym = e == 1 and " " or (e == 2 and " " or "" )
        s = s .. n .. sym
      end
      return s
    end,
....
}
```
Screen Shot 2021-03-24 at 11 12 36 PM
Screen Shot 2021-03-24 at 11 11 53 PM
Screen Shot 2021-03-24 at 11 00 47 PM

P.S. this is my first experience writing Lua, and thank you for the great plugin!

### CODE REVIEW

**Comment by arsaniwilliam** [comment]
sorry didn't squash the commits before the pr.

File: lua/bufferline/diagnostics.lua [hunk 1]
```diff
@@ -1,13 +1,22 @@
 local M = {}
 
--- return the most severe level of diagnostic
-local function get_max_severity(errors)
-  for _, err in ipairs(errors) do
-    if err and err.severity == 1 then
-      return "error"
+local function getErrDict (errs)
```
akinsho, line 3: the convention in lua is to use snake case for identifiers so this should be `get_err_dict`

lua/bufferline/diagnostics.lua
```diff
@@ -1,13 +1,22 @@
...
+      if s < max then max = s end
+      -- increment diagnostics dict
+      if ds[s]
+        then ds[s] = ds[s] + 1
+        else ds[s] = 1
+      end
     end
   end
-  return "warning"
+  local maxSev = max==1 and "error" or (max==2 and "warning" or "info")
```
akinsho, line 18: similar issue as above i.e. snake_case_here

lua/bufferline/diagnostics.lua
```diff
@@ -1,13 +1,22 @@
...
+local function getErrDict (errs)
+  local ds = {}
+  local max = 3
+  for _, err in ipairs (errs) do
+    if err then
+      -- calculate max severity
+      local s = err.severity
+      if s < max then max = s end
+      -- increment diagnostics dict
+      if ds[s]
```
akinsho, line 12: could you reformat this to be
```lua
if ds[s] then
  ds[s] = ds[s] + 1
else
  ds[s] = 1
end
```

lua/bufferline/diagnostics.lua
```diff
@@ -28,10 +37,11 @@ function M.get(opts)
   local diagnostics = vim.lsp.diagnostic.get_all()
   local result = {}
   for buf_num, items in pairs(diagnostics) do
-    result[buf_num] = {
-      count = #items,
-      level = get_max_severity(items)
-    }
+    local d = getErrDict(items)
+    result[buf_num] = { count = #items,
```
akinsho, line 41: sorry more formatting issues, can you change the indent level here so it looks more like it did previously i.e.
```lua
result[buf_num] = {
  count = #items,
  level = d.level,
  errors = d.errors
}
```

akinsho (changes requested):
@arsaniwilliam thanks for the PR this looks good in general. I've made a couple of comments re. casing and formatting. Sorry this project doesn't have an autoformatter in CI or something so it's not clear what the rules are 😓 .

Also I think it'd be good to update the documentation and the README so other people know what is now possible. Maybe you can include some of your screenshots as well so people can get some inspiration 👍🏿

arsaniwilliam:
@akinsho Thank you for the suggestions!
I made the changes you requested, and updated the Readme.

akinsho:
@arsaniwilliam thanks for making the changes, will have a look shortly just a bit busy at the moment but should be free in a little bit

lua/bufferline/diagnostics.lua
```diff
@@ -1,13 +1,23 @@
...
+local function get_err_dict(errs)
+  local ds = {}
+  local max = 3
+  for _, err in ipairs (errs) do
+    if err then
+      -- calculate max severity
+      local s = err.severity
+      if s < max then max = s end
+      -- increment diagnostics dict
+      if ds[s] then
```
akinsho, line 12: @arsaniwilliam I only just noticed/thought of this now but could the structure of this table map to the type of error so that at a glance the object is more legible so users don't have to dig into what `1`,`2` and `3` mean i.e.
```lua
{
 error = 2,
 warning = 0,
 info = 1,
}
```
you could keep an internal mapping of level number to error type rather than exposing the numbers to users which I think is a little less user friendly? WDYT?

akinsho: so basically each object would alway have all three fields but the value related to each type would be the thing that changed.

akinsho: @arsaniwilliam WDYT?

arsaniwilliam: you're right - it's probably more user friendly to return a dict from err_level to count
example: ```lua
{ "error" =3 ,
  "warning" = 5,
  "info" = 2 
}
```
The last commit takes care of that. i'll update the Readme too.

akinsho:
This looks good to merge now @arsaniwilliam thanks for all your work on this 👍🏿

PR #70, opened 2021-04-01
Added new mode 'both'

jerrywang1981:
1. in the new mode 'mix', it will be like lightline-bufferline, with both
   ordinal number and buffer id available.
2. number_style will not work in new mode 'mix'

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ not track global variables which is the mechanism used to store your sort order.
 require'bufferline'.setup{
   options = {
     view = "multiwindow" | "default",
-    numbers = "none" | "ordinal" | "buffer_id",
+    numbers = "none" | "ordinal" | "buffer_id" | "mix",
     number_style = "superscript" | "",
     mappings = true | false,
     buffer_close_icon= '',
--- a/README.md
+++ b/README.md
@@ -34,6 +34,13 @@ see: `:h bufferline-styling`
 
 ![bufferline with numbers](https://user-images.githubusercontent.com/22454918/111993201-3d595600-8b0e-11eb-8944-387ed3bd25b4.png)
 
+mode "both" with default number_style
+![both with default style](https://user-images.githubusercontent.com/8133242/113400253-159ea380-93d4-11eb-822c-974d728a6bcf.png)
+
+mode "both" with customized number_style `{"superscript", "subscript"}`
+![both with customized style](https://user-images.githubusercontent.com/8133242/113400265-1a635780-93d4-11eb-8085-adc328385cb5.png)
+
+
 #### Buffer pick functionality
 
 ![bufferline pick](https://user-images.githubusercontent.com/22454918/111993296-5bbf5180-8b0e-11eb-9ad9-fcf9619436fd.gif)
--- a/README.md
+++ b/README.md
@@ -132,7 +132,7 @@ require'bufferline'.setup{
   options = {
     view = "multiwindow" | "default",
     numbers = "none" | "ordinal" | "buffer_id" | "both",
-    number_style = "superscript" | "",
+    number_style = "superscript" | "" | { "none", "subscript" }, -- buffer_id at index 1, ordinal at index 2
     mappings = true | false,
     buffer_close_icon= '',
     modified_icon = '●',
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ not track global variables which is the mechanism used to store your sort order.
 require'bufferline'.setup{
   options = {
     view = "multiwindow" | "default",
-    numbers = "none" | "ordinal" | "buffer_id" | "mix",
+    numbers = "none" | "ordinal" | "buffer_id" | "both",
     number_style = "superscript" | "",
     mappings = true | false,
     buffer_close_icon= '',
--- a/doc/bufferline-lua.txt
+++ b/doc/bufferline-lua.txt
@@ -49,7 +49,7 @@ The available settings are: >
     require"bufferline".setup{
         options = {
             view = "multiwindow" | "default",
-            numbers = "none" | "ordinal" | "buffer_id",
+            numbers = "none" | "ordinal" | "buffer_id" | "mix",
             number_style = "superscript" | "",
             mappings = true | false,
             buffer_close_icon= "",
@@ -168,7 +168,7 @@ Bufferline can be configured to take a custom filtering function via the
 `custom_filter` option. This value must be a lua function that will receive
 each buffer number that is going to be used for the bufferline, as well as all the others.
 A user can then check whatever they would like and return `true` if they would like it to
-appear and `false` if not. 
+appear and `false` if not.
 For example: >
     custom_filter = function(buf, buf_nums)
         -- dont show help buffers in the bufferline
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -131,7 +131,7 @@ not track global variables which is the mechanism used to store your sort order.
 require'bufferline'.setup{
   options = {
     view = "multiwindow" | "default",
-    numbers = "none" | "ordinal" | "buffer_id",
+    numbers = "none" | "ordinal" | "buffer_id" | "mix",
```
akinsho, line 134: I'd prefer if the new name was `both` not mix. Mix is a little confusing

doc/bufferline-lua.txt
```diff
@@ -49,7 +49,7 @@ The available settings are: >
     require"bufferline".setup{
         options = {
             view = "multiwindow" | "default",
-            numbers = "none" | "ordinal" | "buffer_id",
+            numbers = "none" | "ordinal" | "buffer_id" | "mix",
```
akinsho, line 52: Same here re. naming this `both`

lua/bufferline/numbers.lua
```diff
@@ -26,10 +26,31 @@ local superscript_numbers = {
   [20] = "²⁰"
 }
 
+-- styled numbers at top half
+local t_numbers = { ['0'] = "⁰", ['1'] = "¹", ['2'] = "²", ['3'] = "³", ['4'] = "⁴", ['5'] = "⁵", ['6'] = "⁶", ['7'] = "⁷", ['8'] = "⁸", ['9'] = "⁹" }
```
akinsho, line 30: Can you please collapse this table into multiple lines? same as the one below.
I also think they should be named more clearly. `t_numbers` is actually `superscript_numbers`
and the one below `d_numbers` is `subscript_numbers`

lua/bufferline/numbers.lua
```diff
@@ -26,10 +26,31 @@ local superscript_numbers = {
...
+-- styled numbers at down half
+local d_numbers = { ['0'] = '₀', ['1'] = '₁', ['2'] = '₂', ['3'] = '₃', ['4'] = '₄', ['5'] = '₅', ['6'] = '₆', ['7'] = '₇', ['8'] = '₈', ['9'] = '₉' }
+
+-- from number to the styled number
+local trans_number = function (t, n)
```
akinsho, line 35: I'd rename this function to `convert_to_styled_num` or something that explains what it does

lua/bufferline/numbers.lua
```diff
@@ -26,10 +26,31 @@ local superscript_numbers = {
...
+end
+
 local function prefix(buffer, mode, style)
-  local n = mode == "ordinal" and buffer.ordinal or buffer.id
-  local num = style == "superscript" and superscript_numbers[n] or n .. "."
-  return num
+  -- if mode is mix, it will be like lightline-bufferline, buffer_id at top left
+  -- and ordinal number at bottom down, so the user can get the buffer number
+  if mode == "mix" then
+    return trans_number(t_numbers, buffer.id) .. trans_number(d_numbers, buffer.ordinal)
```
akinsho, line 48: The current appearance of this new mode is *quite* confusing, I think to adjacent superscipt and subscript numbers is not really very clear UI, which is which? how does the user know?

image

akinsho (changes requested):
Thanks for the PR @jerrywang1981, I've left some comments about naming stuff and formatting, but slightly more importantly I think the current way the numbers are laid out is a little confusing. I think there must be a better way to show them maybe something like `1.) filename.js ²` or something

README.md
jerrywang1981: done

doc/bufferline-lua.txt
jerrywang1981: done

lua/bufferline/numbers.lua
jerrywang1981: done

lua/bufferline/numbers.lua
jerrywang1981: done

lua/bufferline/numbers.lua
jerrywang1981: it might be confusing at first look, but with good documentation it will not be problem, and this is how lightline-bufferline looks like with buffer id at left and ordinal number at right. keeping same design might be good to lightline-bufferline users.

screenshot from my previous `lightline-bufferline`
image

akinsho: I prefer the lightline bufferline look to what this PR currently uses which is a mixture of super and subscript numbers. I think it might be nice to still offer the user the option of specifying a style for each type. So maybe when mode is set to `both` then you can specify

```lua
number_style = {id = "superscript", ordinal = "subscript" | "none" | "period" }
-- in numbers.lua
if type(number_style) == "table" and mode == "both" then
  -- apply the right style per section
end
```

So it will be customisable for users. I think the default should be like the screenshot you have above. I don't really like it tbh but I don't use numbers and as long as other users can customise it I'm fine. I think this also needs adding to the docs in help. And maybe a screenshot in the readme with an explanation as well.

akinsho: None can be a full sized number and period `4.` addition of a period to a number

jerrywang1981: not sure where to put the screenshot.

default:
Screen Shot 2021-04-02 at 10 58 12 AM

with 
```
number_style = { "superscript", "subscript"}
```
Screen Shot 2021-04-02 at 11 38 51 AM

akinsho: 👍🏾 this looks great, thanks for tweaking it. Regarding the screenshot you can copy the link from this issue comment and reuse it in the readme.

Or better still there's a closed issue called screenshots where I stash them all. If you post it there and reuse the link then I'll be able to find it better in the future.

jerrywang1981: updated readme

lua/bufferline/numbers.lua
```diff
@@ -3,33 +3,72 @@ local constants = require "bufferline/constants"
...
-  local n = mode == "ordinal" and buffer.ordinal or buffer.id
-  local num = style == "superscript" and superscript_numbers[n] or n .. "."
-  return num
+  -- if mode is both, it will be like lightline-bufferline, buffer_id at top left
+  -- and ordinal number at bottom down, so the user can get the buffer number
+  if mode == "both" then
+    -- default number_style for mode "both"
+    local both_style = { buffer_id = "none", ordinal = "subscript" }
+    if style == "superscript" then -- this is the default number_style, do nothing
+      -- return convert_to_styled_num(superscript_numbers, buffer.id) .. convert_to_styled_num(subscript_numbers, buffer.ordinal)
```
akinsho, line 48: @jerrywang1981 looks like this line is left over now , if you just want to skip updating the `both_style` for you can combine the inverse of the this condition with the one below like.
```lua
	if style ~= "superscript" and type(style) == "table" then
     -- both style updates
    end
```

akinsho:
Just tried this out locally, thanks for tweaking the number style option. It doesn't seem to work in all cases though. I tried `number_style = {"none", "none"} that doesn't give `1. 2.` also tried `{"superscript", "none"}` and that just gives `1.` no super script number

jerrywang1981:
> Just tried this out locally, thanks for tweaking the number style option. It doesn't seem to work in all cases though. I tried `number_style = {"none", "none"} that doesn't give `1. 2.`also tried`{"superscript", "none"}`and that just gives`1.` no super script number

you were right, just fixed  it.

akinsho (approved):
Alright this looks fine to merge now @jerrywang1981 👍🏿

PR #72, opened 2021-04-06
Adding diagnostic shading separate from component shading

Reptarsrage:
I liked the fact that in vscode the diagnostic indicators inside of the tabs are a shade darker, so I added the same functionality here!


lua/bufferline.lua
```diff
@@ -114,30 +114,43 @@ local function get_buffer_highlight(buffer, hls)
...
     hl.info = h.info_visible.hl
+    hl.info_diagnostic = h.info_diagnostic_visible.hl
   else
     hl.background = h.background.hl
     hl.modified = h.modified.hl
     hl.duplicate = h.duplicate.hl
     hl.pick = h.pick.hl
     hl.separator = h.separator.hl
     hl.buffer = h.background
+    hl.diagnostic = h.diagnostic.hl
```
akinsho, line 145: What does the diagnostic hl on it's own cover?

I see the buffer text highlight is separate to the diagnostic highlight now so two highlights are needed which is fine. But not sure about the case of `diagnostic.hl` seems like this is when the severity isn't one of the known ones?

lua/bufferline/diagnostics.lua
```diff
@@ -73,19 +73,23 @@ function M.component(context)
...
     return context.component, context.length
   end
+
   local indicator = " (" .. diagnostics.count .. ")"
   if user_indicator and type(user_indicator) == "function" then
     indicator = user_indicator(diagnostics.count, diagnostics.level, diagnostics.errors)
   end
+
   local highlight = highlights[diagnostics.level] or ""
+  local diag_highlight = highlights[diagnostics.level.."_diagnostic"] or highlights["diagnostic"] or ""
```
akinsho, line 90: `highlights["diagnostics"]` can just be `highlights.diagnostics`

lua/bufferline/colors.lua
```diff
@@ -1,5 +1,667 @@
 local M = {}
 
+-- X11 color names taken from
+-- https://en.wikipedia.org/wiki/X11_color_names
+local x11_color_names = {
```
akinsho, line 5: I'm not sure I want to take this on, i.e. a giant mapping of colors built into the plugin itself. I think users can override almost any colour this plugin sets with whatever they'd like so I don't think this needs to happen here. It also doesn't appear to be related to this PR's core functionality

akinsho (changes requested):
@Reptarsrage thanks for the PR 👍🏿 I've left a couple of comments, the biggest/only blocking issue I have is with the change to `colors.lua`. I'm generally good with the change.

I personally like it but am a little wary of follow up issues if people prefer the previous behaviour but I'd rather not add the complexity of yet another option, so people can set the color themselves if they don't like it.

Side note: do you have a screenshot of the vscode behaviour this is now similar to? I don't use it so wasn't aware of it doing this

lua/bufferline/colors.lua
Reptarsrage: The issue this is trying to solve is really with the `get_hex` function. `get_hex` doesn't work as expected when given resolved color names like "Red" and "Orange". These color names were being returned by `vim.fn.synIDattr` instead of the expected hex values. I needed these color names to be hex values so that I could apply `shade_color` to them. I don't like having this list either but I couldn't figure out another way to get hex color codes from named colors.

lua/bufferline.lua
Reptarsrage: Yeah when you want diagnostic indicators on a tab when the tab isn't currently in info/warning/error state. I personally don't want this option, but it seems like a good thing to have in case people want to display  diagnostic indicators in this case.

Reptarsrage:
> Side note: do you have a screenshot of the vscode behaviour this is now similar to? I don't use it so wasn't aware of it doing this

here's a screenshot

lua/bufferline/diagnostics.lua
Reptarsrage: Good call. Fixed.

lua/bufferline.lua
Reptarsrage: and more specifically they want to control the color of the diagnostic indicators in this case

lua/bufferline/colors.lua
Reptarsrage: Perhaps there's some swanky vim function I'm not aware of which given "Red" can return "#FF0000", but I coudn't track it down

akinsho: Ahh that is an interesting case, I'm curious where is this happening? is your colour scheme using explicit colours in the `gui` attributes. I can think of a few different solutions one would be checking for hex via a regex and using the fallback if not. A better solution would be migrating to `nvim_get_hl_by_name('Value', true)` I've just checked and this consistently returns an rgb value.

To use it for highlight commands though would need a function to convert to and from `rgb` something like
```javascript
function RGBToHex(r,g,b) {
  r = r.toString(16);
  g = g.toString(16);
  b = b.toString(16);

  if (r.length == 1)
    r = "0" + r;
  if (g.length == 1)
    g = "0" + g;
  if (b.length == 1)
    b = "0" + b;

  return "#" + r + g + b;
}
```

So all colours can be pulled out as rgb but converted to hex for usage, primarily because that's what the highlight commands use. `nvim_set_hl` doesn't interact with the normal `highlight` space yet.

In either case I'd rather this problem be resolved in a separate PR since it isn't directly related to anything in this particular PR though

lua/bufferline.lua
akinsho: Makes sense 👍🏿

lua/bufferline/colors.lua
Reptarsrage: I'm not sure what's causing `vim.fn.synIDattr(vim.fn.hlID("Error"), "fg")` to return "Red". Maybe it's my theme? But if it's happening to me we should probably consider handling it. Adding this change without a fix for that will break the plugin.

I'm noticing that `vim.api.nvim_get_hl_by_name("Red", true)` throws error "Invalid highlight name"

Reptarsrage: I'll spend some time gonig through the code and replacing `vim.fn.synIDattr` with `vim.api.nvim_get_hl_by_name` and we can get that merged before merging this

Reptarsrage: Or would you rather merge this change first instead? I'll leave that up to you

akinsho: I'd rather merge this first without the colour changes as I'm not sure that this change specifically introduces this issue since it doesn't actually add new highlights and the `Error` issues you referenced isn't new to this issue. I think if a user's colour scheme is returning a colour name for any of this plugins highlights it's been an issue since day one of this plugin since I've used this mechanism since then.

Also you shouldn't have to search for multiple places to replace `synIDattr` all colour fetching happens via `colors.get_hex` so just changing that function should be enough

Reptarsrage: okay the color changes have been removed :+1:

akinsho (approved):
LGTM 🚀 thanks again for the contribution

PR #76, opened 2021-04-06
Adding named color support in get_hex

Reptarsrage:
get_hex will now correctly parse out named colors such as "Red" and "Orange" as well formatted hex color codes!

### DIFF
--- a/lua/bufferline/colors.lua
+++ b/lua/bufferline/colors.lua
@@ -60,15 +60,16 @@ function M.color_is_bright(hex)
 end
 
 function M.get_hex(hl_name, part, fallback)
-  local id = vim.fn.hlID(hl_name)
-  local color = vim.fn.synIDattr(id, part)
+  local color = fallback
+  local hl = vim.api.nvim_get_hl_by_name(hl_name, true)
 
-  -- TODO: synIDattr may return a named color such as "Red"
-  -- we should use vim.api.nvim_get_hl_by_name(hl_name, true)
+  -- translate from internal part to hl part
+  assert(part == "fg" or part == "bg", 'Color part should be one of "fg" or "bg"')
+  part = (part == "fg" and "foreground") or "background"
 
-  -- if we can't find the color we default to none
-  if not color or color == "" then
-    return fallback
+  if hl and hl[part] then
+    -- convert from decimal color value to hex (e.g. 14257292 => #D98C8C)
+    color = string.format("#%x",  hl[part])
   end
 
   return color


Reptarsrage:
NOTE: [nvim docs](https://neovim.io/doc/user/api.html) say that `nvim_get_hl_by_name({name}, true)` should return RGB but it returns hex color codes in decimal form 🤷🏻

lua/bufferline/colors.lua
```diff
@@ -60,15 +60,18 @@ function M.color_is_bright(hex)
...
-  local id = vim.fn.hlID(hl_name)
-  local color = vim.fn.synIDattr(id, part)
+  local color = fallback
+  local hl = vim.api.nvim_get_hl_by_name(hl_name, true)
 
-  -- TODO: synIDattr may return a named color such as "Red"
-  -- we should use vim.api.nvim_get_hl_by_name(hl_name, true)
+  -- translate from internal part to hl part
+  assert(part == "fg" or part == "bg", 'Color part should be one of "fg" or "bg"')
+  part = (part == "fg" and "foreground") or "background"
```
akinsho, line 68: I think this could just be
```lua
-- no need for the parens
local attrbute = part == "fg" and "foreground" or "background"
```

lua/bufferline/colors.lua
```diff
@@ -60,15 +60,18 @@ function M.color_is_bright(hex)
...
 
-  -- if we can't find the color we default to none
-  if not color or color == "" then
-    return fallback
+  if hl and hl[part] then
+    -- convert from decimal color value to hex (e.g. 14257292 => #D98C8C)
+    color = string.format("%x",  hl[part])
```
akinsho, line 72: I *think* that through some magic in lua courtesy of c you can maybe do
```lua
color = string.format("%06x", hl[part])
```
Reading through the lua guide [here](https://www.lua.org/pil/20.html) right at the bottom I think he format characters can specify a length and a character to use for padding. It talks about decimal in that example but think it might work here

lua/bufferline/config.lua
```diff
@@ -9,13 +9,13 @@ function M.get_defaults()
   local hex = colors.get_hex
   local shade = colors.shade_color
 
-  local comment_fg = hex("Comment", "fg", "NONE")
-  local normal_fg = hex("Normal", "fg", "NONE")
+  local comment_fg = hex("Comment", "fg", "#D3D3D3")
```
akinsho, line 12: As a rule I avoid *any* hardcoded hex colours because people's color schemes and tastes are far too varied. 

I'm guessing the issue here is that the named colours and `NONE` colours don't play well with `shade_color` in which case I'd recommend finding a native highlight group to link them to that looks sensible.

I think if the `hl` is `NONE` shade color should just return the `NONE` i.e. if it can't shade a color or the color isn't a hex it should just return the thing as is.

In this specific case I think these should look more like
```lua
local warning_fg = hex("LspDiagnosticsDefaultWarning", "fg", {name = "WarningMsg", attribute = "fg"})
```
Then the implementation of `get_hex` should try to resolve the first hl group and if it fails try to return the fallback and if that fails then return `NONE`

lua/bufferline/colors.lua
```diff
@@ -60,15 +60,18 @@ function M.color_is_bright(hex)
 end
 
 function M.get_hex(hl_name, part, fallback)
-  local id = vim.fn.hlID(hl_name)
-  local color = vim.fn.synIDattr(id, part)
+  local color = fallback
```
akinsho, line 63: The fallback shouldn't be expected to be a hex colour because as a rule this plugin should never be specifying specific hex colours. I think fallback should either be something that can be passed directly to `guifg` or `guibg` like a color name or `NONE`. or as I mention below should be another highlight group to use in case the first one doesn't return a colour

akinsho (changes requested):
@Reptarsrage 🚀 . Thanks for picking this up so quickly. I've made a few comments. One around maybe a nice way to handle the padding of the the hex, if it works. The other around hardcoding hex values which I try to avoid at all costs and think can be reworked to just use some other core highlight groups as fallbacks instead

lua/bufferline/colors.lua
Reptarsrage: man I wish lua had ternary operators 😅

Reptarsrage: fixed.

lua/bufferline/colors.lua
Reptarsrage: That's slick as hell 🐱‍👤

Reptarsrage: fixed.

lua/bufferline/colors.lua
Reptarsrage: get_hex now takes a table as fallback which will chain recursive calls. "NONE" is now the default fallback

lua/bufferline/config.lua
Reptarsrage: implemented.

akinsho (approved):
This looks great @Reptarsrage thanks for your work on this 👍🏿 .
Weird about the nvim docs versus the return values... hope there's  nothing weird going on

PR #90, opened 2021-05-06
feat: add offset for sidebar windows

### LINKED ISSUE #89: bufferline length
is It possible to change bufferline length ? I dont want it 100% and get mixed with NvimTree (file tree on the left)

image

I want it to get separated from the NvimTree , something like this 

image

akinsho:
image

<details>

image

</details>

This PR fixes #89 by adding an offset to the bufferline so it begins (and eventually) ends where a sidebar panel starts or ends. This gives the impression that the bufferline is local to the window, similar to GUI editors.

The implementation is based on adding padding/text before or after the bufferline so is somewhat hacky, if this somehow proves problematic in the long term this might need to be revisited +/- deprecated.

### TODO
- [x] Add documentation
- [x] Handle windows open on the right
- [x] Add tests

NOTE: I am *not* taking more feature requests on top on this functionality, I'm not going to add anything jazzier to this i.e. it will be text or an empty space above the window

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,7 +174,7 @@ require'bufferline'.setup{
         return true
       end
     end,
-    panels = {{filetype = "NvimTree", text = "File Explorer"}},
+    offsets = {{filetype = "NvimTree", text = "File Explorer"}},
     show_buffer_close_icons = true | false,
     show_close_icon = true | false,
     show_tab_indicators = true | false,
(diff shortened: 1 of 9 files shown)


siduck76:
@akinsho Thanks a lot! is working soo good :D 

image

PR #97, opened 2021-05-13
feat: add recent sort option

akinsho:
Allow the buffers to be sorted based on how frequently a user enters that particular buffer. Not super sure how useful this will turn out to be.


akinsho:
close in favour of #151

PR #99, opened 2021-05-15
Get icon from terminal key

ram02z:
I use nvim-nonicons, which has a few overrides from the original nvim-web-devicons plugins. One of which is terminal, which has the decimal value of [62601](https://github.com/kyazdani42/nvim-web-devicons/blob/1db27380053de0cd4aaabd236a67c52d33199f1a/lua/nvim-web-devicons.lua#L856) in the original plugin, whilst the nonicons plugin has the decimal value of [61911](https://github.com/yamatsum/nvim-nonicons/blob/5056aebb8d0ecc87b26a4f7fe9e9e520e5ea493f/lua/nvim-nonicons/mapping.lua#L216) for the same icon.
This results in a broken glyph, which could also be a result of my bad font patch. Either way, as long as the key is called `"terminal"`, this should fix it.

### DIFF
--- a/lua/bufferline/buffers.lua
+++ b/lua/bufferline/buffers.lua
@@ -45,6 +45,9 @@ function M.Buffer:new(buf)
   buf.extension = fn.fnamemodify(buf.path, ":e")
   -- Set icon
   if buffer_is_terminal(buf) then
+    if lua_devicons_loaded then
+      terminal_icon = webdev_icons.get_icon("terminal")
+    end
     buf.icon = terminal_icon
     buf.filename = fn.fnamemodify(buf.path, ":p:t")
   else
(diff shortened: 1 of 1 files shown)


lua/bufferline/buffers.lua
```diff
@@ -45,6 +45,9 @@ function M.Buffer:new(buf)
   buf.extension = fn.fnamemodify(buf.path, ":e")
   -- Set icon
   if buffer_is_terminal(buf) then
+    if lua_devicons_loaded then
+      terminal_icon = webdev_icons.get_icon("terminal")
```
akinsho, line 49: @ram02z there's actually some slightly buggy behaviour I've noticed which is that this was reliant on there being a space in the icon I included in this plugin. Can you append a space here otherwise the icon gets mangled.

akinsho (changes requested):
Thanks for the PR @ram02z this looks good generally just one small change is needed

ram02z:
> 
> 
> Thanks for the PR @ram02z this looks good generally just one small change is needed

No problem. Added the small change in latest commit.

akinsho (approved):
Thanks for tweaking this LGTM

PR #100, opened 2021-05-15
feat: add ability to add a custom section

### LINKED ISSUE #98: Feature request - area for custom content
Hey 👋🏾 

your plugin looks very appealing and has really nice features. 👍🏾 

I really like to have such a nice tab-styled bufferline. But I also want to have some additional global information shown there. For example I would like to try moving the git branch info from the statusline to the bufferline because it won't change per buffer/window but is a more global information (for my use-cases). Another example would be the status of the language server or number of global workspace diagnostics. I actually would love to have a unified or second statusline at the bottom of NeoVim, but that does not exist.
I don't want to ask you to support all these features. Rather I was thinking if it wouldn't be a nice generic feature to let the user have a reserved space on the right side of the bufferline for his own content. So there would be two options. One for the guaranteed space this segment gets and one for a function which returns the content of this segment. Interesting would be to have different autocommand events which would trigger to re-evaluate the function to update the segment, but that's maybe too much at once. The buffer "tabs" should then just truncate as now. Maybe the guaranteed size is not even necessary and could be more dynamic. But that sounds more complicated and would make a first version of this feature way harder (I assume).

What do you think about this idea in general? 🙃 
I think it is kinda cool, because it allows to use your fancy UI parts and combine if a quite generic interface to become extended with some additional useful information. Looking forward to hear what you have to say. 🤗

akinsho:
This will allow a user to specify a custom section on either end of the statusline.
This can be used as
```lua
config = {
 options = {
	custom_areas = {
		left = function()
	 		return {
       			text = "  " .. vim.lsp.diagnostic.get_count(0, "Error"),
       			guifg = "salmon",
     		},
		end
		}
 	}
}
```
image

Fixes #98, cc @weilbith can you give this a try and see if it works for you. I really don't want to go down any rabbit holes having to add refresh logic or anything so if it doesn't I might have to scrap this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,6 @@ not track global variables which is the mechanism used to store your sort order.
 ```lua
 require('bufferline').setup {
   options = {
-    view = "multiwindow" | "default",
     numbers = "none" | "ordinal" | "buffer_id" | "both",
     number_style = "superscript" | "" | { "none", "subscript" }, -- buffer_id at index 1, ordinal at index 2
     mappings = true | false,
--- a/README.md
+++ b/README.md
@@ -197,13 +197,6 @@ require('bufferline').setup {
 }
 ```
 
-#### NOTE:
-
-If using a plugin such as `vim-rooter` and you want to sort by path, prefer using `directory` rather than
-`relative_directory`. Relative directory works by ordering relative paths first, however if you move from
-project to project and vim switches its directory, the bufferline will re-order itself as a different set of
-buffers will now be relative.
-
 ### LSP Error indicators
 
 By setting `diagnostics = "nvim_lsp"` you will get an indicator in the bufferline for a given tab if it has any errors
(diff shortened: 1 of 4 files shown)


akinsho:
@elianiva I think I noticed in some screenshots of your config somewhere (compe's readme I think) that you use something like this for your bufferline can you try out this implementation and see if it works for you?

elianiva:
great work! I'm able to replicate my previous setup with this much shorter and better config.

<details>

<summary> config </summary>

```lua
  custom_areas = {
    right = function()
      local result = {}
      local error = vim.lsp.diagnostic.get_count(0, [[Error]])
      local warning = vim.lsp.diagnostic.get_count(0, [[Warning]])
      local info = vim.lsp.diagnostic.get_count(0, [[Information]])
      local hint = vim.lsp.diagnostic.get_count(0, [[Hint]])

      if error ~= 0 then
        result[1] = { text = "  " .. error, guifg = "#EC5241" }
      end

      if warning ~= 0 then
        result[2] = { text = "  " .. warning, guifg = "#EFB839" }
      end

      if hint ~= 0 then
        result[3] = { text = "  " .. hint, guifg = "#A3BA5E" }
      end

      if info ~= 0 then
        result[4] = { text = "  " .. info, guifg = "#7EA9A7" }
      end

      return result
    end
  },
```

</details>

akinsho:
@elianiva if you don't mind I might use your example in the docs so people can see a potential use case for a custom area

elianiva:
sure! feel free to use it :)

PR #103, opened 2021-05-21
Add missing commas to config example

bapjiws:


akinsho (approved):
Thanks, hopefully people aren't copy pasting this gigantic block directly since it's not intended to be valid for that usage but still doesn't hurt to have commas in the right place

PR #109, opened 2021-05-25
Fix: incorrect padding of hex color parts

### LINKED ISSUE #107: Colors looks different after d6c226e8e05155101b202d7c94e1f9d1e4b2d145
Before:
изображение

After:
изображение

After this change, it became more difficult to distinguish between open buffers because of vertical white line on each buffer. Also, this line after the current buffer is moved to the right after changing the current buffer.

Is this expected? Can I somehow get previous behavior?

akinsho:
@Shatur95 can you try this PR to see if it works for you? I didn't/don't notice a difference on my machine so it must be the shade of your colorscheme.

Will fix #107


Shatur95:
Yes, it fixes the issue, thanks!

PR #117, opened 2021-06-08
feat: add configurable mouse actions

### LINKED ISSUE #106: Allow using custom close function / command
So as you might already know, the default `:bdelete` command in Neovim is quite annoying in the sense that it messed up the window layouts. I made a function that bypasses it and would want to be able to configure nvim-bufferline to use it instead of `:bdelete` or whatever else it uses for deleting that buffer. So it would be really nice to have that option.

akinsho:
This PR adds a `close_command` which can be a function or string. If a function it is passed the buffer id and if a string it will have the buffer id appended (this needs thinking since it's less versatile)
Fixes #106 

@famiu can you test this branch out and see if it works for you.


famiu:
Yup, it works for me

PR #124, opened 2021-06-13
feature: add commands to close all buffers in a direction

akinsho:
Adds `CloseBufferLineRight` and `CloseBufferLineLeft` to close all buffers to the right or left of the current buffer.


akinsho:
fixes #124

Issue #130, opened 2021-06-26
[Feature request] custom names


pwntester:
Hi!

I was wondering if it could be possible to allow changing the name appearing in the tabline items. My use case is that if the buffer name starts with `octo:/` I would like to show a text with a truncated version of the github issue that the buffer represents.

Thanks
A

akinsho:
Hi @pwntester,

This isn't too complicated a feature to add, and It could have some cool use cases. What concerns me is that people are almost certainly going to abuse this to change things other than the sort of use case you describe, which will very probably break the plugin and result in issues like "why can't I make x name work" etc. I guess this could be solved by stern documentation, I'm just wary of opening the floodgates to a range of issues regarding buffer names.

akinsho:
Closed by #131

pwntester:
Thanks!

PR #131, opened 2021-06-27
feature: add name formatter

akinsho:
This PR adds the ability to format the name a buffer is given in the bufferline. It is intended to potentially make otherwise confusing names clearer or for minor tweaks to naming.

It is not intended as a way for users to use full paths as names or inject highlighting etc. If users use it for such it will break, and I will not fix that 🤷🏾‍♂️ .

The key can be specified as
```lua
--- NOTE that if nil is returned the default will be used anyway
name_formatter = function(default_name, path)
  if condition then
    return <new-name>
  end
end
```

### Todo
- [x] Tests
- [x] Sterner wording/warning docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,16 @@ require('bufferline').setup {
     close_icon = '',
     left_trunc_marker = '',
     right_trunc_marker = '',
+    --- name_formatter can be used to change the buffer's label in the bufferline.
+    --- Please note some names can/will break the
+    --- bufferline so use this at your discretion knowing that it has
+    --- some limitations that will *NOT* be fixed.
+    name_formatter = function(buf)  -- buf contains a "name", "path" and "bufnr"
+      -- remove extension from markdown files for example
+      if buf.name:match('%.md') then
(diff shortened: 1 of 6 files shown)


akinsho:
@pwntester can you give this a try to see if it fits your use case?

akinsho:
fixes #130

pwntester:
Thanks for implementing this and for doing it so fast!

Im using it like this:

```lua
      name_formatter = function(_, path)
        local bufnr = vim.fn.bufnr(path)
        if vim.startswith(path, "octo:") and octo_buffers and octo_buffers[bufnr] then
          local title = octo_buffers[bufnr].node.title
          return string.sub(title, 1, 10)
        end
      end,
```

Not sure if I can always trust the output of `vim.fn.bufnr(path)` though. Can you pass the bufnr as well?

akinsho:
@pwntester I've tweaked this now so that the function returns a table containing a `name`, `path` and `bufnr`. Let me know if that works for you and I'll tidy this up and merge it.

pwntester:
Awesome, thanks! works like a charm!

PR #134, opened 2021-06-30
feature: add tabs sorter 

stevanmilic:


akinsho:
@stevanmilic thanks for the PR, will get to reviewing it soon. In the meantime can I ask you to also add a test for the new sorter. You should only have to test the specific function you added in `sorters_spec.lua` (please make one if I haven't already 🤦🏾 ) there you can import your function and just validate that it does what you expect.

Regarding tests you can see the other ones for some examples or check `plenary.nvim`'s testing module which is what this uses and that has some docs or `busted` which is the lua framework that that copies.

lua/bufferline/sorters.lua
```diff
@@ -40,6 +40,38 @@ local function sort_by_directory(buf_a, buf_b)
...
+
+  for _, tab in ipairs(tabs) do
+      local buffers = vim.fn.tabpagebuflist(tab.tabnr)
+      for _, buffer_id in ipairs(buffers) do
+          if buffer_id == buffer_a.id then
+              buffer_a_tabnr = tab.tabnr
+          elseif buffer_id == buffer_b.id then
+              buffer_b_tabnr = tab.tabnr
+          end
+          if buffer_a_tabnr ~= nil and buffer_b_tabnr ~= nil then
```
akinsho, line 58: You don't have to check explicitly against `nil` these checks can be `if not buffer_a_tabnr and not buffer_b_tabnr then`

lua/bufferline/sorters.lua
```diff
@@ -40,6 +40,38 @@ local function sort_by_directory(buf_a, buf_b)
   return full_path(buf_a.path) < full_path(buf_b.path)
 end
 
+--- @param buf_a Buffer
```
akinsho, line 43: The parameter names must match the function names, otherwise the lsp won't pick these up correctly. In this case, your parameters should be called `buf_a`, `buf_b`.

akinsho (changes requested):
@stevanmilic I tried this out, and I'm actually getting an error.
```
E5108: Error executing lua ...art/local-nvim-bufferline.lua/lua/bufferline/sorters.lua:86: invalid order function for sorting
```
I also made a few minor comments

stevanmilic:
@akinsho I'll definitely write some tests, this PR is still in WIP phase. But I wanted you to take a look. 

Yea, I'm having the same errors, for me the problem occurs when there are hidden buffers that don't have a tab number, then when a buffer that is null is compared in the order function it could produce different sorting output. Probably because it _shifts_ places in the parameter list. I'm still playing with the function output when one of the buffer is null, an acceptable output is to have hidden buffers at the end or beginning of the buffer list. Not sure how I can write test for this, but I'll try.

lua/bufferline/sorters.lua
stevanmilic: you mean `if buffer_a_tabnr and buffer_b_tabnr then`, right?

akinsho:
@stevanmilic if a buffer is hidden and not visible in any tab, why not default the value to 0 which is essentially neutral or whatever value will stick them on the end since if you had 30 hidden buffers, and you stuck them at the start you'd likely never see the sorting this function provides.

Regarding testing, you can start small with testing the function, plenary's test framework allows you to mock the result of things like `vim.fn.funcname` or `vim.api.nvim_func` so you can just make those return what you need for the test.

stevanmilic:
@akinsho I've made some changes to the sorting function, can you please check if everything looks good at your end? Now the hidden buffers should be placed at the end of the list.

lua/bufferline/sorters.lua
```diff
@@ -40,6 +40,42 @@ local function sort_by_directory(buf_a, buf_b)
...
+--- @param buf_b Buffer
+local function sort_by_tabs(buf_a, buf_b)
+  local tabs = vim.fn.gettabinfo()
+  local buf_a_tabnr = nil
```
akinsho, line 47: These should probably default to 0 so there are never any `nil`s

stevanmilic: Well if it's a 0 then all the hidden buffers would be stacked at the beginning of the list. We should use `math.maxinteger` but it looks like it's added from lua 5.3. I can just use some custom-defined max int value, then it should be good.

akinsho: I think some arbitrarily high number is fine if it's commented 🤷🏾 if there is a more formal thing that'd be great though, although could you invert the comparison?

akinsho:
@stevanmilic going to be busy for the next few days but will try and have a look early next week

akinsho:
@stevanmilic just tried this out briefly in my lunch break and it doesn't seem to work how I expected it to. Say I opened a few buffers `a-e` then focus `a` in tab1 then go to tab two and split it in two and show `e` and `d` then go to three and show `c` I'd expect the order to be `a,e,d,c,b` but it isn't

stevanmilic:
Hmm yea I can reproduce it, for some reason the buffer `d` can't be found from the `tabpagebuflist`. Adding another buffer in a new tab, then correctly sorts the list as if the `d` appears _again_.

lua/bufferline/sorters.lua
stevanmilic: Inverting came to my mind too, but it shouldn't be possible since returning `true` with sorting functions would place `buf_a` in front of `buf_b`, so just by inverting the comparison we gain nothing because `0` would still be lower than any number.

stevanmilic:
@akinsho I managed to fix the problem you reported, I found out the buffer gets hidden (?!) when a buffer in last tab is focused, and it's not found through `tabpagebuflist`. But in all other cases it _is_ visible, so the solution was to **not** use the `maxinteger` default value for all the buffers. Instead we use the `maxinteger` value for hidden buffers only. Interestingly the `buffer:visible()` function that you've written also reports the buffer being visible, but in fact it isn't, using `next(vim.fn.win_findbuf(buf.id)) ~= nil` condition correctly recognizes the buffer as hidden.

I think that it's best to re-write the https://github.com/akinsho/nvim-bufferline.lua/blob/165114f65c267ee912018ef3565d70cba7183fc3/lua/bufferline/buffers.lua#L96 to use the same _correct_ expression that I wrote. Check the first comment on https://stackoverflow.com/a/10219740/2413731.

lua/bufferline/sorters.lua
```diff
@@ -44,6 +44,43 @@ end
 --- @param buf_b Buffer
 local function sort_by_id(buf_a, buf_b)
   return buf_a.id < buf_b.id
```
akinsho, line 46: I believe this is a syntax error, the function doesn't have an `end` so it breaks the plugin

akinsho (changes requested):
@stevanmilic nice find with the fix, I believe there's a syntax error here though from a missing end keyword. Probably from a bad merge. Also I noticed whilst using it that things aren't always in order i.e. I opened a buffer in tab 1 and 2 windows in tab 2 with 2 other buffers and the window in tab 1 wasn't the first in the list. should the order be windows in tab1 first then those in tab 2 then all the hidden?

lua/bufferline/sorters.lua
stevanmilic: ye, I'll fix that real quick 👍🏻

stevanmilic:
@akinsho I didn't manage to re-produce what you just reported. The buffer in tab 1 always stays first in the list, what does happen is that in tab 2 windows could have a different order in regards to how they are split -- the left window could be _after_ the right window in the list, but I don't think it we should try to resolve that.

akinsho:
@stevanmilic just tried again and this appears to work fine now 👍🏿

akinsho (approved):
LGTM 🚀

stevanmilic:
@akinsho I'll write some tests, and then we can merge? I see that you have written a `sorters_spec`, so I can take the ideas from that 👍🏻

akinsho:
Sounds great @stevanmilic lemme know if you need a hand with those. The other tests should help, but also plenary has some docs on how it's testing framework works. It's basically copied from busted.

stevanmilic:
@akinsho looks like mocking the `vim.fn` isn't possible https://github.com/nvim-lua/plenary.nvim/issues/166 :|

akinsho:
@stevanmilic that's a shame, although you could potentially just reproduce the correct scenario in the nvim e.g. 
````lua
	vim.cmd [[
      e file1.txt
      tabnew file2.txt
      tabnew file3.txt
    ]]
```
and then maybe just let the actual functions run? if that still doesn't work then we can just abandon the attempt.

stevanmilic:
@akinsho I have added a test similar to what you proposed, but for some reason the test ain't passing. I can't understand why that happens, maybe I'm using a wrong function to retrieve bufer IDs? Could you please take a look?

akinsho:
@stevanmilic thanks for your perseverance on getting the tests working. I had a look locally and I think the issue is that the sorters actually expect `---@type Buffer` rather than just the buffer number.

I managed to get it working with
```diff
diff --git a/tests/minimal.vim b/tests/minimal.vim
index 7547e39..7b5043c 100644
--- a/tests/minimal.vim
+++ b/tests/minimal.vim
@@ -1,3 +1,4 @@
 set rtp+=.
 set rtp+=../plenary.nvim
 runtime! plugin/plenary.vim
+set noswapfile
diff --git a/tests/sorters_spec.lua b/tests/sorters_spec.lua
index 026b2cf..dd45648 100644
--- a/tests/sorters_spec.lua
+++ b/tests/sorters_spec.lua
@@ -15,13 +15,15 @@ describe("Sorters - ", function()
     vim.cmd("tabnew file2.txt")
     vim.cmd("tabnew file3.txt")
     vim.cmd("bunload file2.txt")
-    local bufs = vim.api.nvim_list_bufs()
+    local bufs = vim.tbl_map(function(id)
+      return { id = id }
+    end, vim.api.nvim_list_bufs())
 
     sorters.sort_buffers("tabs", bufs)
 
-    local buf_names = vim.tbl_map(function(buf_num)
-      return vim.fn.fnamemodify(vim.api.nvim_buf_get_name(buf_num), ":p:t")
+    local buf_names = vim.tbl_map(function(buf)
+      return vim.fn.fnamemodify(vim.api.nvim_buf_get_name(buf.id), ":p:t")
     end, bufs)
-    assert.same({ "file1.txt", "file3.txt", "file2.txt"}, buf_names)
+    assert.same({ "file1.txt", "file3.txt", "file2.txt" }, buf_names)
   end)
 end)
```

stevanmilic:
@akinsho I've made the changes you suggested, thanks for the input! But still the test doesn't pass in my local environment? Let's try on CI?

lua/bufferline/sorters.lua
```diff
@@ -46,6 +46,44 @@ local function sort_by_id(buf_a, buf_b)
...
 end
 
+--- @param buf_a Buffer
+--- @param buf_b Buffer
+local function sort_by_tabs(buf_a, buf_b)
+  local maxinteger = 1000000000
+
+  local init_value = function(buf)
```
akinsho, line 54: Not a huge deal but maybe this can be factored out, so it isn't redeclared on every call to this function which might happen 100s of times 🤷🏿‍♂️

akinsho (approved):
LGTM 🎉 thanks for your work on this will merge this in half an hour or so if you're done 👍🏿

akinsho:
@stevanmilic regarding running things locally you need to make sure it is running with the `minimal.vim` otherwise it will run the tests with your nvim config which will muddy the waters

lua/bufferline/sorters.lua
stevanmilic: ye sure, I've added there for readability.

PR #136, opened 2021-07-04
Typo in README.md s/sessionopts/sessionoptions/

IndianBoy42:
the setting should be `sessionoptions` not `sessionopts`, atleast on neovim 0.5.0. 

I don't have vim installed to check but googling seems to show its sessionoptions there too


akinsho:
Thanks 👍🏾

Issue #146, opened 2021-07-17
After latest commit, bufferline is not being hidden when `showtabline=0` while on dashboard.


Pocco81:
Hello @akinsho! The tittle says it all. This started happening since updating the plugin two days ago. I read through the code and [removing `VimEnter`](https://github.com/akinsho/nvim-bufferline.lua/commit/cebafb95622205a414a6c10bf0e40d197cc652b1#diff-47fabfe452ff7a153117c6d9b99f2ef62455cb8ae325dd77fd62d85fe3d10d4dR889) from the event seems to be the problem.

Could you add it back and leave it as it was before?

akinsho:
Thanks for reporting this @Pocco81 I don't use a startup menu and didn't notice a difference in testing so I assumed it was serving no purpose. Does it work as expected for you if the event is re-added?

Pocco81:
Surprisingly enough, adding VimEnter didn't work. However I did revert to the commit 344564a and everything works as expected. Are the changed from the latest commit necessary?

akinsho:
@Pocco81 that's not really necessary should be able to fix it without resorting to such drastic measures

Pocco81:
What can we/you do then?

akinsho:
@Pocco81 I understand what is probably going on and have raised a PR can you try it and report back on there if it fixes the issue.

Pocco81:
Nope, it's still the same :(

This is my config btw:
```
require "bufferline".setup {
    options = {
        mappings = true,
        enforce_regular_tabs = true,
        buffer_close_icon = "",
        modified_icon = "●",
        close_icon = "",
        left_trunc_marker = "",
        right_trunc_marker = "",
        numbers = "ordinal",
        number_style = "superscript",
        diagnostics = "false",
        max_name_length = 14,
        max_prefix_length = 13,
        tab_size = 16,
        view = "multiwindow",
        show_buffer_close_icons = true,
        separator_style = "thin",
        offsets = {{filetype = "NvimTree", text = "Explorer"}},
        always_show_bufferline = false -- hide buffer if there are less than one
    }
    -- highlights = {
    -- 	background = {					-- inactive tab color
    -- 		guifg = comment_fg,
    -- 		guibg = "#282c34"
    -- 	},
    -- 	buffer_selected = {				-- active tab color
    -- 		guifg = normal_fg,
    -- 		guibg = "#3A3E44",
    -- 		gui = "bold"
    -- 	},
[...]

PR #147, opened 2021-07-17
fix: hide bufferline on start if not configured

akinsho:
This PR changes startup to only show the bufferline on startup if a user has configured it to show up

@Pocco81 please try this to see if it fixes the issue.

### DIFF
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -2,6 +2,7 @@ local constants = require("bufferline.constants")
 local utils = require("bufferline.utils")
 
 local api = vim.api
+local fn = vim.fn
 local fmt = string.format
 -- string.len counts number of bytes and so the unicode icons are counted
 -- larger than their display width. So we use nvim's strwidth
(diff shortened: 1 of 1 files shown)


akinsho:
should fix #146

akinsho:
@Pocco81 I don't see any bufferline when I use this PR using dashboard. If you are using packer, did you make sure to reload the config or restart your neovim?

Pocco81:
Whoops! My bad! since the change in this PR was just one line I simply copy and pasted it locally. Turns out I pasted it in the wrong place. Yes, now everything works :)

akinsho:
@Pocco81 I thought of a bug so just repushed a new solution, can you cross-check it still works for you and if so I'll merge it.

Pocco81:
Yep, it works as well. Also, I noticed that you call `vim.cmd` and `vim.o` a lot of times. Maybe you should consider instantiate both like you did with `vim.fn` for example? (I mean: `local cmd = vim.cmd). That would make the code more readable :+1:

Issue #150, opened 2021-07-18
Can I ues BufferLinePick to close a buffer?


KeqiZeng:
Hello bro, I don't use mouse in neovim, so I want to know can I use the function `BufferLinePick` to close a buffer just like pick it. If yes, can you tell me how to get it?
Thank you so much!!

akinsho:
@KeqiZeng that isn't possible currently, it's an interesting idea though. I'm probably not going to look into it myself anytime soon, but would appreciate/review a PR for it though

ramezgerges:
@akinsho I gave it a try (#153).

akinsho:
This is fixed by #153 which added the `BufferLinePickClose` command

KeqiZeng:
OK, I'll try it. Thank you bro!

PR #153, opened 2021-07-19
feature: add command to select and close a buffer

ramezgerges:
This commit adds the functionality requested in #150. Namely,  a command to pick a buffer and close it (similar to BufferLinePick but closes instead of selects). I've also refactored bufferline.pick_buffer() so that both it and the new bufferline.close_buffer() can share the same logic. I've refrained from adding anything to README.md as my writing style would probably clash with @akinsho 's. Also my color scheme is a bit different so pictures would look out of place. I've tried adding a test or two but I couldn't find any tests for bufferline.pick_buffer(), so I could not figure out how to give user input in code.

P.S. this is my first pull request ever, please be gentle.

### DIFF
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -52,22 +52,6 @@ function M.restore_positions()
   end
 end
 
-function M.pick_buffer()
-  state.is_picking = true
-  refresh()
-
-  local char = vim.fn.getchar()
-  local letter = vim.fn.nr2char(char)
-  for _, buf in pairs(state.buffers) do
-    if letter == buf.letter then
-      vim.cmd("buffer " .. buf.id)
-    end
-  end
-
-  state.is_picking = false
-  refresh()
-end
-
 ---Handle a user "command" which can be a string or a function
 ---@param command string|function
 ---@param buf_id string
@@ -108,6 +92,36 @@ function M.handle_click(id, button)
   if id then
     handle_user_command(options[cmds[button]], id)
   end
+end
+
+-- Prompts user to select a buffer then applies a function to the buffer
+-- @param func function
+function select_buffer_apply(func)
+  state.is_picking = true
+  refresh()
+
+  local char = vim.fn.getchar()
+  local letter = vim.fn.nr2char(char)
+  for _, buf in pairs(state.buffers) do
+    if letter == buf.letter then
+      func(buf.id)
+    end
+  end
+
+  state.is_picking = false
+  refresh()
+end
+
+function M.pick_buffer()
+  select_buffer_apply(function(buf_id)
+    vim.cmd("buffer " .. buf_id)
+  end)
+end
+
+function M.close_buffer()
+  select_buffer_apply(function(buf_id)
+    M.handle_close_buffer(buf_id)
+  end)
 end
 
 ---@param buffer Buffer
@@ -914,6 +928,7 @@ end
 local function setup_commands()
   local cmds = {
     { name = "BufferLinePick", cmd = "pick_buffer()" },
+    { name = "BufferLineClose", cmd = "close_buffer()" },
     { name = "BufferLineCycleNext", cmd = "cycle(1)" },
     { name = "BufferLineCyclePrev", cmd = "cycle(-1)" },
     { name = "BufferLineCloseRight", cmd = 'close_in_direction("right")' },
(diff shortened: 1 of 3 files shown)


lua/bufferline.lua
```diff
@@ -110,6 +94,36 @@ function M.handle_click(id, button)
   end
 end
 
+-- Prompts user to select a buffer then applies a function to the buffer
+-- @param func function
```
akinsho, line 98: You can actually make this more specific so the lsp knows more about the function that is expected, e.g.

```lua
---@param func fun(id: number)
```
If you then hover over func it will have the correct type signature, it's a minor point in this case though since it's a simple function but could warn someone if they tried to use it incorrectly.

lua/bufferline.lua
```diff
@@ -914,6 +928,7 @@ end
 local function setup_commands()
   local cmds = {
     { name = "BufferLinePick", cmd = "pick_buffer()" },
+    { name = "BufferLineClose", cmd = "close_buffer()" },
```
akinsho, line 931: I'd call this `BufferlinePickClose` or something since close doesn't specify the mechanism, if there were ever any other more generic close command then this name would be taken as well. Wish I'd have named the original command pick open or something but hadn't really planned for this and now it would be a breaking change.

lua/bufferline.lua
```diff
@@ -110,6 +94,36 @@ function M.handle_click(id, button)
...
+  refresh()
+end
+
+function M.pick_buffer()
+  select_buffer_apply(function(buf_id)
+    vim.cmd("buffer " .. buf_id)
+  end)
+end
+
+function M.close_buffer()
```
akinsho, line 121: I'd also rename this, so it's clear that it relates to the pick functionality. `close_buffer` in the context of this, plugin is a pretty generic name for this and I don't think it'd be clear that this used the/was for the pick functionality

akinsho: maybe `close_buffer_with_pick` or something 😅 not entirely sure.

akinsho (commented):
@ramezgerges I like this approach a lot. Thanks for picking this up. I had been thinking it would be a good solution to generalise this function, so the pick functionality could be used to do other things, and this step allows for that. I'd also been thinking being able to have a different highlight for each type of pick might be nice, but think that's a whole other chunk of work (that doesn't need to happen here).

Regarding tests, you can pass input to nvim using `nvim_feedkeys` I added the test infra after this plugin had existed for a while, so there's still a lot of untested regions. I wouldn't consider it a blocker but if you wanted to have a go using feedkeys it should theoretically work but as I said not a blocker.

Regarding docs, tbh those are a bit of a pain that I otherwise will forget to update for sure otherwise. I don't think it necessarily needs to have a picture or GIF but just some simple mention of this functionality and what it does would be good, I can always tweak the wording later.

Overall this looks pretty good and tested it out so will be happy to merge I'd just want to make the name a little less general, so somewhere it mentions that it involves picking and a doc update 🙏🏿

lua/bufferline.lua
ramezgerges: So, turns out I kept being lazy and forgot to setup LSP for lua. I've just set that up and fixed the parameter thing. I totally agree on the name. I've actually spend quite a bit of time thinking of an appropriate name and couldn't find anything, so I'll fix that now.  I will take a look at the README, shouldn't be too hard. I have no clue whatsoever about testing in lua, what should I google for? Is this a lua thing or a vim thing? looks extremely similar to those js testing frameworks.

akinsho: So the testing framework plenary basically copied and gutted is [busted](http://olivinelabs.com/busted/#matchers), but it didn't copy ever method there. Another place to look would be plenary's README [here](https://github.com/nvim-lua/plenary.nvim/blob/master/TESTS_README.md). I wouldn't sink loads of time into the tests, there are a few gotchas, but I think this should hopefully not be too bad.

ramezgerges:
@akinsho Let me know what you think about the last 2 commits. Nothing earth-shattering, so I guess there aren't many things that can go wrong. I tried testing with Plenary and I was way out of my depth. Maybe in a future PR.

akinsho:
@ramezgerges this looks good to go now I've added in some docs for the help text as well. Will merge this in now 👍🏿 . Thanks again for your work on this.

PR #156, opened 2021-07-21
Fix Vimscript usage in README

alexandregv:
Hi, a little PR to fix the Vimscript usage in the README:
* The vimscript comment must be outside the lua heredoc, otherwise it triggers an error
* No `lua` keyword required before the `require()` (also triggers an error)

image


akinsho:
Thanks for the fix @alexandregv 👍🏿

PR #158, opened 2021-07-21
Refactor: improve get_updated_buffers

ramezgerges:
I saw this TODO in the code and thought I could do something about it. The original implementation took quadratic time (O(|sorted|*|buf_nums|)). This one is linear. This makes use of a utility function tbl_reverse_lookup, which is similar to vim.tbl_add_reverse_lookup, except that this one creates a new table instead of modifying the original (this is necessary because the both the keys and the values are numerical, so you can't exchange in place). This implementation reorders buf_nums in place. I've checked the references and they're all in the form of "buf_nums =  get_updated_buffers(buf_nums, sorted)", so I assume it's fine to do it in place. If not, I'm not sure exactly how to shallow copy buf_nums ellegantly (i.e. without a loop) as table.unpack is now depreciated and I'm not sure if there is a current alternative.

### DIFF
--- a/lua/bufferline/utils.lua
+++ b/lua/bufferline/utils.lua
@@ -59,15 +59,14 @@ end
 --- creates a table whose keys are tbl's values and the value of these keys
 --- is their key in tbl (similar to vim.tbl_add_reverse_lookup)
 --- this assumes that the values in tbl are unique and hashable (no nil/NaN)
---- @param tbl table
---- @return table
+--- @generic K,V
+--- @param tbl table<K,V>
+--- @return table<V,K>
 function M.tbl_reverse_lookup(tbl)
     local ret = {}
-
-    for i, b in pairs(tbl) do
-      ret[b] = i
+    for k, v in pairs(tbl) do
+      ret[v] = k
     end
-
     return ret
 end
 
--- a/lua/bufferline/utils.lua
+++ b/lua/bufferline/utils.lua
@@ -58,13 +58,13 @@ end
 
 --- creates a table whose keys are tbl's values and the value of these keys
 --- is their key in tbl (similar to vim.tbl_add_reverse_lookup)
---- this assumes that the values in tbl are unique
+--- this assumes that the values in tbl are unique and hashable (no nil/NaN)
 --- @param tbl table
 --- @return table
 function M.tbl_reverse_lookup(tbl)
     local ret = {}
 
-    for i, b in ipairs(tbl) do
+    for i, b in pairs(tbl) do
       ret[b] = i
     end
 
--- a/lua/bufferline/utils.lua
+++ b/lua/bufferline/utils.lua
@@ -56,6 +56,21 @@ function M.filter_duplicates(array)
   return res
 end
 
+--- creates a table whose keys are tbl's values and the value of these keys
+--- is their key in tbl (similar to vim.tbl_add_reverse_lookup)
+--- this assumes that the values in tbl are unique
+--- @param tbl table
+--- @return table
+function M.tbl_reverse_lookup(tbl)
+    local ret = {}
+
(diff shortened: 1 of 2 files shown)


lua/bufferline/utils.lua
```diff
@@ -56,6 +56,21 @@ function M.filter_duplicates(array)
...
 end
 
+--- creates a table whose keys are tbl's values and the value of these keys
+--- is their key in tbl (similar to vim.tbl_add_reverse_lookup)
+--- this assumes that the values in tbl are unique and hashable (no nil/NaN)
+--- @param tbl table
+--- @return table
+function M.tbl_reverse_lookup(tbl)
+    local ret = {}
+
```
akinsho, line 66: Can you remove the empty lines here, so this is a bit more compact

lua/bufferline/utils.lua
```diff
@@ -56,6 +56,21 @@ function M.filter_duplicates(array)
   return res
 end
 
+--- creates a table whose keys are tbl's values and the value of these keys
+--- is their key in tbl (similar to vim.tbl_add_reverse_lookup)
+--- this assumes that the values in tbl are unique and hashable (no nil/NaN)
+--- @param tbl table
```
akinsho, line 62: You can type annotate a table, so it can be `---@param tbl table<number, number>` etc. so the lsp knows the types to expect

akinsho (approved):
This looks great, thanks for working on this @ramezgerges. Regarding copying a table, you can use `vim.deepcopy()` but not sure if it's a huge deal. I'll give this a try locally, but generally looks good to me

lua/bufferline/utils.lua
ramezgerges: I think this works with pretty much any key/value types, as long as the values are unique (so that keys are unique when reversed). Is there something like table<T, T> like generics in C++/Java?

lua/bufferline/utils.lua
ramezgerges: I'm not sure if you want me to copy the buf_nums or if you just wanted to inform me that vim.deepcopy() is a thing. Should I use it here or not?

lua/bufferline/utils.lua
akinsho: Yeah there is actually a
```lua
---@generic T
---@param tbl table<T,T>
```
See: https://github.com/sumneko/lua-language-server/wiki/EmmyLua-Annotations#generic

lua/bufferline/utils.lua
akinsho: Sorry @ramezgerges I didn't look at things in the greatest detail so was only mentioning that there was mechanism for copying if that was what you'd have opted to do had you known. At a glance I don't think it matters.

ramezgerges:
@akinsho the max_integer thing was bugging me so I cleaned that up. Also fixed the annotations and the spaces.

akinsho:
Gave this a quick try and seems to work fine on my end, thanks for the contribution @ramezgerges 👍🏿

ramezgerges:
@akinsho sorry to bring it up again, but now that I've thought about it, it's ridiculously unwise to not copy buf_nums. The problem is not that it modifies the original, the problem is that it returns a reference to the original, so one might think they are manipulating a new copy when in fact they aren't. It's a one liner tho (plus removing the comment about it being in place), so I don't think I need to open a new PR for it. Proceed however you see fit.

akinsho:
Thanks for the heads up @ramezgerges, I personally always prefer not mutating data since it just makes things harder to follow. I'll tweak this when next I get a chance. It isn't dire yet though as tbh it's passed a newly created lifted of ids and the life span of this list is pretty short i.e. it's not being used in a bunch of different places but that could always change

Issue #166, opened 2021-07-26
[Feature] Dynamic offset text


Akianonymus:
Currently, the offset text at top is static, fetched during initialisation and then it uses the same.

How about using a dynamic text ? Adding something like offset_type option, which can be string or function

If offset_type is string, then show as it is, otherwise evaluate the string ( using loadstring ) and use it.

siduck76:
I think it already does 

`local current_cwd = vim.fn.fnamemodify(vim.fn.fnamemodify.getcwd(), ':t');

 text = current_cwd
`

Akianonymus:
@siduck76 No it doesn't, obviously i checked with a function before creating this issue.

It won't because it doesn't evaluate it further, it gets what it gets at the first time.

Also, the function i was using.

```
function nvim_tree_cwd()
    local cwd = vim.fn.getcwd()
    local tree_loaded, lib = pcall(require, "nvim-tree.lib")
    if tree_loaded then
        cwd = lib.Tree.cwd
    end
    local root_folder_modifier = vim.g.nvim_tree_root_folder_modifier or ":t"
    cwd = vim.fn.fnamemodify(cwd, root_folder_modifier)
    local final_cwd = cwd

    if cwd and cwd ~= "/" then
        if tree_loaded then
            local utils = require "nvim-tree.utils"
            final_cwd = utils.path_join(utils.path_remove_trailing(cwd), "..")
        end
    end
    return final_cwd
end
```

siduck76:
> @siduck76 No it doesn't, obviously i checked with a function before creating this issue.
> 
> It won't because it doesn't evaluate it further, it gets what it gets at the first time.
> 
> Also, the function i was using.
> 
> ```
> function nvim_tree_cwd()
>     local cwd = vim.fn.getcwd()
>     local tree_loaded, lib = pcall(require, "nvim-tree.lib")
>     if tree_loaded then
>         cwd = lib.Tree.cwd
>     end
>     local root_folder_modifier = vim.g.nvim_tree_root_folder_modifier or ":t"
>     cwd = vim.fn.fnamemodify(cwd, root_folder_modifier)
>     local final_cwd = cwd
> 
>     if cwd and cwd ~= "/" then
>         if tree_loaded then
>             local utils = require "nvim-tree.utils"
>             final_cwd = utils.path_join(utils.path_remove_trailing(cwd), "..")
>         end
>     end
>     return final_cwd
> end
> ```
> 
> 

[...]

PR #170, opened 2021-07-27
offset.lua: Allow using a function for offset text

Akianonymus:


lua/bufferline/offset.lua
```diff
@@ -15,6 +15,9 @@ local t = {
 ---@return string
 local function get_section_text(size, highlight, offset)
   local text = offset.text
+  if offset.text_type == "function" then
+    text = _G[text]()
+  end
```
akinsho, line 20: @Akianonymus this doesn't need to be specified as a separate field, and use lua's `type` argument to check the type of the field. I suggest
```suggestion
if type(text) == "function" then
 text = text()
end
```

akinsho (changes requested):
@Akianonymus just made a suggestion to improve how this should work so that extra fields and manual type specification by a user isn't required.

lua/bufferline/offset.lua
akinsho: There's also no need to reference the global object in lua since that will point at a non-existent text field and cause errors.

akinsho:
Regarding docs you can update the reference to offset where it specifies that text is a `string` argument to show that it is now a union of string or function e.g. `string | function` and explain that the function will be called to generate a string each time.

Akianonymus:
@akinsho Please check, i have pushed the changes.

akinsho:
LGTM @Akianonymus thanks for the PR

akinsho:
fixes #166

PR #180, opened 2021-08-10
fix: truncate multibyte characters by cell

### LINKED ISSUE #177: [Bug]: Abnormal Chinese display in the bufferline.
### What happened?

截屏-20210808151940-980x1068

It's expected that the titles are displayed correctly.

### Additional Information

I'm not clear about the reason, but I guess that you may need the function `vim.fn.strwidth` which returns the size of string when displayed.

### commit

_No response_

akinsho:
This PR adds trunctation of multibyte strings by cell/column rather than by byte, only if the byte count of a string does not match display size i.e. if it contains multibyte characters

@alohaia can you try this PR and see if it fixes #177

### DIFF
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -216,6 +216,24 @@ local function get_buffers_by_mode(mode)
   return utils.get_valid_buffers()
 end
 
+-- truncate a string based on number of display columns/cells it occupies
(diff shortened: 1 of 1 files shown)


akinsho:
closes #177

alohaia:
Yes, it works perfectly now. Thanks for your patience and time.

PR #184, opened 2021-08-15
feature: added ordinal_first mode to numbers

### LINKED ISSUE #178: [Feature request] Add option to swap `ordinal` and `buffer_id` position with `numbers = "both"`
I like the result when `numbers = 'both'` and `number_style = { 'none', 'superscript' }` are set:

` 1.¹   foo |  3,²   bar |  4.³ baz |`

But I see `ordinal` as more useful than `buffer_id`, so instead I would like the above to be: 

` 1.¹   foo |  2,³   bar |  3.⁴ baz |`

Would you consider implementing something like that?

I60R:
refactor: related code in numbers.lua and README.md
docs: document ordinal_first in README.md
closes #178

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,8 +182,8 @@ not track global variables which is the mechanism used to store your sort order.
 ```lua
 require('bufferline').setup {
   options = {
-    numbers = "none" | "ordinal" | "buffer_id" | "both",
-    number_style = "superscript" | "" | { "none", "subscript" }, -- buffer_id at index 1, ordinal at index 2
+    numbers = "none" | "ordinal" | "buffer_id" | "both" | "ordinal_first",
+    number_style = "none" | "superscript" | "subscript" | { "none", "subscript" }, -- buffer_id at #1, ordinal at #2
     mappings = true | false,
     close_command = "bdelete! %d",       -- can be a string | function, see "Mouse actions"
     right_mouse_command = "bdelete! %d", -- can be a string | function, see "Mouse actions"
--- a/lua/bufferline/config.lua
+++ b/lua/bufferline/config.lua
@@ -387,7 +387,7 @@ local function get_defaults()
     options = {
       view = "default",
       numbers = "none",
-      number_style = "superscript",
+      number_style = nil,
       buffer_close_icon = "",
       modified_icon = "●",
       close_icon = "",
--- a/lua/bufferline/numbers.lua
+++ b/lua/bufferline/numbers.lua
@@ -31,45 +31,49 @@ local subscript_numbers = {
 -- from number to the styled number
 local convert_to_styled_num = function(t, n)
   n = tostring(n)
-  local r = ""
+  local str = ""
   for i = 1, #n do
-    r = r .. t[n:sub(i, i)]
+    str = str .. t[n:sub(i, i)]
   end
-  return r
+  return str
 end
 
+-- build numbers string depending on config settings
 local function prefix(buffer, mode, style)
-  -- if mode is both, it numbers will look similar lightline-bufferline, buffer_id at top left
-  -- and ordinal number at bottom right, so the user see the buffer number
-  if mode == "both" then
-    -- default number_style for mode "both"
-    local both_style = { buffer_id = "none", ordinal = "subscript" }
-    if style ~= "superscript" and type(style) == "table" then
-      both_style.buffer_id = style[1] and style[1] or both_style.buffer_id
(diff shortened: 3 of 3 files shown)


akinsho:
@I60R thanks for the PR, will take a proper look in a bit 👍🏿

akinsho:
@I60R I wonder if this PR could address https://github.com/akinsho/nvim-bufferline.lua/issues/178#issuecomment-902303555 by actually adding not another option but just the ability for a user to specify a function, so they can format the numbers themselves. Something like
```lua
number_style = function(ordinal, buf_id)
	return ordinal .. " + " .. buf_id
end
```
This way there are a few builtin options but when a user wants more granular control they can tweak things however they like and the plugin doesn't have to support a growing list of styles to accomodate a bunch of different preferences.

I60R:
With this PR the list of styles seems to be complete and probably wouldn't grow further.

I'd like it to be just a function if it would have been implemented initially in that way, but at this point IMO breaking backwards compatibility doesn't make any sense. The list of options is also sensible enough, so with it users may not need a formatting function.

---

It's a good candidate for change if you would issue a `v2` of this plugin

akinsho:
@I60R 
> With this PR the list of styles seems to be complete and probably wouldn't grow further.

I highly doubt this, unfortunately. Across all my plugins I have always assumed x number of options must *surely* be enough, tbh even in the case of this option I already think the current list of options is "enough" but someone will always want something else which I won’t always provide, but this will definitely change without the function option.

To be clear making this *also* take a function is *not* a breaking change. There are a few options already that either take a string *or* a function. In this case my suggestion is rather than introducing a new `ordinal_first` which tbh isn't really a new style but a preference on an old style I would just add in a new function option and users can just do whatever they want without me needing to continously add new styles e.g.

```
number_style = 'both' | 'orginal' | {<thing>, <thing>} | function(ordinal, buf_id)
```

then in the code we can do
```lua
if type(preferences.number_style) == "function" then
 return preferences.number_style(buf.ordinal, buf.id)
end
-- everything below is the way it was already
```

I60R:
You may forgot that currently there are two options:

```lua
numbers = "none" | "ordinal" | "buffer_id" | "both",
number_style = "superscript" | "" | { "none", "subscript" }, -- buffer_id at index 1, ordinal at index 2
```

Unfortunately, I don't see a clear and backward-compatible way to add function to any of them

I60R:
Ideally there should have been something like

```lua
numbers = nil | function(ordinal, buffer_id, lower, raise)
    -- add custom formatting
   return '#' .. lower(ordinal) .. raise(buffer_id) .. gap  -->  #₅⁴²   buffername.ext
end,
```

akinsho:
@I60R there's nothing preventing us adding the numbers function option you suggested above, if the numbers option is specified we can just ignore the numbers style option. I also think that all the other styles can be implemented in terms of the function above. I also think the numbers style can then basically just be a proxy for writing some number functions on the user's behalf. In the future, I might deprecate the numbers style and require users who want that granularity to write a numbers function or could also just allow `_subscript` or `_superscript` version of the numbers enum.


Regarding the current API I did not envision allowing so much flexibility and I should have said no when the earlier requests to change the formatting were made. I started with 2 options but got a never ending series of request to tweak x,y,z.

Going forward I will probably reject further changes as they just make things more complex. I think people get overly specific about this kind of minutiae, and it just makes things less maintainable.

I did not start with a function because a lot of nvim users are still quite new to lua and asking them to write their own function to get the desired result in most cases is more trouble than it's worth.

I60R:
>I did not start with a function because a lot of nvim users are still quite new to lua and asking them to write their own function to get the desired result in most cases is more trouble than it's worth.

You might provide a set of predefined functions e.g. [like author of specs.nvim did](https://github.com/edluffy/specs.nvim#usage). I think in this way users will have either sensible defaults and enough flexibility, and the code would be as well a lot more simpler and easier to modify.

```lua
local bufferline = require('bufferline')
bufferline.setup {
  options = {
    numbers = bufferline.numbers_style.ordinal_subscript,
    sort_by = bufferline.buffer_sort.directory,
    -- etc.
```

This is just suggestion, I'm not trying to push anything.

Regardless this PR, it's fine to ignore it as the solution here is really suboptimal and I certainly could live with forked version. If you would decide that adding custom formatter function is still better — that's just three lines of code that shouldn't depend on it.

Already too much time is spent here...

PR #194, opened 2021-08-22
feature: more tab indicator options

ihasdapie:
Add option & support for showing buffer name, tabnr, both, or provide a custom function for tab indicator text. 

image

Not sure how this will impact offset/spacing calculation, etc. But some qualitative testing shows that it works well enough.


README.md
```diff
@@ -235,6 +235,8 @@ require('bufferline').setup {
     show_buffer_close_icons = true | false,
     show_close_icon = true | false,
     show_tab_indicators = true | false,
+    tab_indicator_option = 'both' | 'tabnr' | 'title' | function(str)
```
akinsho, line 238: I think `tab_indicator_style` would be a better name for this rather than `option`

lua/bufferline/config.lua
```diff
@@ -421,7 +422,8 @@ local function get_defaults()
       show_buffer_icons = true,
       show_buffer_close_icons = true,
       show_close_icon = true,
-      show_tab_indicators = true,
+      show_tab_indicators = false,
+      tab_indicator_option = 'both',
```
akinsho, line 426: I don't think we should change the default behaviour. I think this new style will result in a bit of a duplication of buffer naming and take up space, so should be something someone opts into

lua/bufferline/tabs.lua
```diff
@@ -9,13 +9,30 @@ local function tab_click_component(num)
...
+
   local h = highlights
   local hl = is_active and h.tab_selected.hl or h.tab.hl
   local separator_hl = is_active and h.separator_selected.hl or h.separator.hl
   local separator_component = style == "thick" and "▐" or "▕"
   local separator = separator_hl .. separator_component
-  local name = padding .. padding .. tab.tabnr .. padding
+
+  -- TODO: There's bound to be a better way to get the buffer name!
+  local bufname = vim.fn.getbufinfo(
```
akinsho, line 21: I don't think this should be done here. Information for a tab is derived in the `get` function rather than the render, which is just about the visual appearance of the derived data.

akinsho: You can add a key to the tab table in the `get` function called `name` and
do the name derivation there. I think you can also simplify this to
```lua
local buf = api.nvim_win_get_buf(tab.windows[1])
local name = fn.fnamemodify(api.nvim_buf_get_name(name), ':t')
````

File: lua/bufferline/tabs.lua [hunk 4]
```diff
@@ -9,13 +9,30 @@ local function tab_click_component(num)
...
+      vim.fn.getwininfo(tab.windows[1])[1].bufnr
+      )[1].name
+
+  if tab_indicator_style == 'tabnr' or (bufname == '' or not bufname) then
```
akinsho, line 25: I don't think the name should default to just a number if the `bufname` is empty, I think only if the style is tabnr should there only be a number. You can default this to something like `[no name]`. Basically the `(bufname == '')` clause shouldn't be there, I think if using `both` or `title` and the bufname is `''` then it should be defaulted the same way as is done for buffers

lua/bufferline/tabs.lua
```diff
@@ -9,13 +9,30 @@ local function tab_click_component(num)
...
+    bufname = tab.tabnr
+  elseif tab_indicator_style == 'title' then
+    bufname = bufname:match("^.+/(.+)$")
```
akinsho, line 28: I think this is attempting to do the same thing as `fn.fnamemodify(name, ":t")` this is a more standard and battle tested way built into vim to handle shortening a filename. I've actually added it to the tab name derivation bit above so it doesn't have to be repeated twice here for both styles

lua/bufferline/tabs.lua
```diff
@@ -9,13 +9,30 @@ local function tab_click_component(num)
...
+  elseif tab_indicator_style == 'both' then
+    bufname = tab.tabnr .. ': '  .. bufname:match("^.+/(.+)$")
```
akinsho, line 30: I think this separator should be more easily customizable, as I can immediately see wanting to put a different symbol other than the colon there

lua/bufferline/tabs.lua
```diff
@@ -9,13 +9,30 @@ local function tab_click_component(num)
...
+  else
+    bufname = tab_indicator_style(bufname)
+  end
+
+  local name = padding .. padding .. bufname  .. padding
   local length = strwidth(name) + strwidth(separator_component)
```
akinsho, line 36: As your PR description alludes to, the `length` *has* to be incremented to accommodate these new changes since the name is what has changed, and it's length is being checked here this is why it is likely working. It'd be good to verify this if you can just log out the value locally to make sure it matches what you are expecting 🙏🏿

akinsho (changes requested):
@ihasdapie thanks a lot for the PR 🙏🏿 🚀 it's much appreciated. I've made some comments about the implementation, I'm generally fine with the whole idea and only had a few things that I think could be tweaked. I get a lot of tabs plz issues that I've been considering how/if to accommodate, and this looks like a nice step.

README.md
akinsho: Forgot to mention I think this should be advertised more prominently in the README and also updated in the help docs as well since hopefully it will reduce some "I want to see tabs" issues I get

ihasdapie:
@akinsho Your comments look great, I'll go make edits later today. I think `tab_ indicator_style` can be a table then?

One feature I was considering adding was custom tab names, so something that you can map to a keybinding that'll prompt the user for a tab name string. Thoughts?

lua/bufferline/tabs.lua
ihasdapie: Yeah that looks better than what I had. This still has the problem of displaying the buffer name of the _first_ window in the tabpage, not the name of the _last active_ window which would probably be more desirable

akinsho:
@ihasdapie I don't think `tab_indicator_style` needs to be a table, it can just be  string enum of `'both' | 'tabnr' | 'title' | function` just the change of `option` to `style` since the config doesn't use the naming `option` for any other keys. `style` is the convention for changing the appearance of something I've used so far.

Regarding changing the name, I've seen some other plugins do this. It's a bummer it doesn't reference anything that actually exists in vim, so the name is fake/not related to anything in vim proper, but I guess the name could be added by the user and then stored in the tab object till they next reload vim. Later on if needed the names can be saved using uppercase `vim.g` variables because those are persisted across vim sessions.

ihasdapie:
@akinsho  Ah I was thinking of having it be a table because we wanted to customize the separator? Or just another option would be enough.

akinsho:
Ah I see, I think another option would be enough. Can't envision how the table work

ihasdapie:
No time this week or so, hopefully I can finish up the changes next week maybe

Issue #192, opened 2021-08-21
[Feature Request]: Option for padding between buffer icon and buffer name


HumblePresent:
### What happened?

I recently started using [nonicons](https://github.com/yamatsum/nvim-nonicons) which are slightly larger than the typical icons from nerd fonts. For that reason, I include an extra space after each icon in my config for [nvim-web-devicons](https://github.com/kyazdani42/nvim-web-devicons) so that there is sufficient gap between the icon and any text after it. With this config, bufferline displays the icon followed by 2 spaces and then the buffer name, exactly as I would like. The issue comes from buffer pick mode where the unique letter generated for each buffer is displayed in place of the buffer icon with only one space between it and the buffer name. Going from two spaces to one space for each buffer causes the whole bufferline to visually shrink when entering buffer pick mode and then expand after selecting a buffer. I'm wondering if it's possible to add an option for a number of padding spaces between the buffer icon and buffer name that would be respected by buffer pick mode. 
### What did you expect to happen?

In my case this would result in the buffer pick letter displayed with two spaces between it and the buffer name to keep the width of each buffer the same.

### Additional Information

...

akinsho:
@HumblePresent I think this plugin is getting very weighed down with innumerable configurations and options making the codebase increasingly hard to work with. This particular option is one I'd rather not expose.

Tbh I'm not sure why nonicons would need more than one space since it should just take as many cells as needed for example multibyte characters so not require extra padding. Could it be how your terminal is handling rendering them

HumblePresent:
@akinsho The icons themselves are still rendered in two cells like glyphs from other icon or patched fonts. I have added an extra third cell of space because the glyphs from nonicons render a bit larger within those two cells than other fonts. It just keeps the icon from appearing too close to the adjacent text. That said I totally understand your desire to keep the number of configuration parameters down and not expose so many small details. This is more of a wish list feature anyway. It's not gonna keep me from using the plugin or anything. Thanks for your consideration though.

PR #198, opened 2021-08-25
Add icon padding in buffer pick mode

HumblePresent:
This PR addresses the issue I described in #192 but without exposing another obscure configuration option. Basically, I added what I am calling "icon padding" when entering buffer pick mode. This is extra padding after the buffer letter that matches the width of the buffer icon so that the component won't change width when an icon that is more than a single character is replaced by a single letter. I included some short videos of the before and after in case my explanation of the issue isn't clear. I realize this is a bit of an obscure use case so I understand if you prefer to not include this change, but given it's a very small code change I thought I would see how you felt.

### Without Padding
https://user-images.githubusercontent.com/60856003/130717983-f261150a-617e-4511-9a91-907ddf8d44b4.mp4

### With Icon Padding
https://user-images.githubusercontent.com/60856003/130717921-ae8fc3f9-e9d3-4cf0-b841-d7f69561b91c.mp4

### DIFF
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -562,8 +562,12 @@ local function add_prefix(context)
   local length = context.length
 
   if state.is_picking and buffer.letter then
-    component = hl.pick .. buffer.letter .. padding .. hl.background .. component
-    length = length + strwidth(buffer.letter) + strwidth(padding)
+    local icon_padding = ""
+    if options.show_buffer_icons and buffer.icon then
+      icon_padding = string.rep(" ", strwidth(buffer.icon) - 1)
+    end
+    component = hl.pick .. buffer.letter .. icon_padding .. padding .. hl.background .. component
+    length = length + strwidth(buffer.letter) + strwidth(icon_padding) + strwidth(padding)
   elseif options.show_buffer_icons and buffer.icon then
     local icon_highlight = highlight_icon(buffer)
     component = icon_highlight .. hl.background .. component


lua/bufferline.lua
```diff
@@ -562,8 +562,12 @@ local function add_prefix(context)
   local length = context.length
 
   if state.is_picking and buffer.letter then
-    component = hl.pick .. buffer.letter .. padding .. hl.background .. component
-    length = length + strwidth(buffer.letter) + strwidth(padding)
+    local icon_padding = ""
+    if options.show_buffer_icons and buffer.icon then
+      icon_padding = string.rep(" ", strwidth(buffer.icon) - 1)
```
akinsho, line 567: I try to avoid using a bare empty string where possible and instead use `require('bufferline.constants').padding` it's the same thing but I like that I can control all the padding in the plugin with one variable.

lua/bufferline.lua
```diff
@@ -562,8 +562,12 @@ local function add_prefix(context)
...
+    end
+    component = hl.pick .. buffer.letter .. icon_padding .. padding .. hl.background .. component
```
akinsho, line 569: Rather than just adding padding *after* the letter it seems to me like the letter should be surrounded by the padding e.g. `half_padding .. letter .. half_padding`

akinsho (commented):
@HumblePresent thanks for looking into a fix for this I guess anyone who uses multibyte icons would encounter this issue granted nonicons is the only place where this happens for now so I'm fine to include something like this.

Just left 2 comments about adding the padding around the icon rather than just to the right of it.

Also can you verify that this doesn't break for normal devicon users

lua/bufferline.lua
```diff
@@ -562,8 +562,12 @@ local function add_prefix(context)
   local length = context.length
 
   if state.is_picking and buffer.letter then
-    component = hl.pick .. buffer.letter .. padding .. hl.background .. component
-    length = length + strwidth(buffer.letter) + strwidth(padding)
+    local icon_padding = ""
+    if options.show_buffer_icons and buffer.icon then
```
akinsho, line 566: I think this can be a ternary e.g. 
```lua
local icon_padding = options.show_buffer_icons and buffer.icon and string.rep() or ""
```

Issue #160, opened 2021-07-23
[Feature request] Stop updating lsp diagnostics count when auto completing.


moshiur-raj:
Bufferline updates the lsp diagnostics count when selecting a suggestion from the auto completion menu. I think this is unnecessary and distracting.
It would be much better if bufferline updated the count only when editing in normal mode or when entering the normal mode from insert mode. This is also the default behavior in `nvim-lspconfig`.

I have a question....
Does turning this option on cause neovim to check the file for errors twice? once for bufferline and then for the lsp diagnostic messages? or does neovim only do it once and reuse the diagnostic message?

akinsho:
@moshiur-raj the plugin uses the exposed diagnostics information, similar to how you would use it in your statusline. It doesn't trigger new requests or cause a whole bunch of logic to run. The diagnostics module just saves this information for things like plugins/users to read.

Regarding not showing updates under certain conditions, it might or might not be easy to add it could just be a question of wrapping the function in a mode check or it could involve needing to add autocommand to specifically redraw the tabline, in either case happy to review a PR for it.

moshiur-raj:
I see.
Neovim 0.5 is really awesome and has made me very interested in learning lua. It's waaay better than vimscript.
I'll have to give it a go then...

akinsho:
I believe this can be closed by #209 which added an option to not update diagnostics in insert mode, you can have a look at that PR for more details.

moshiur-raj:
the problem still persists though. is there an option i need to enable?

akinsho:
@moshiur-raj yes there is please check the linked PR for details

moshiur-raj:
thanks,  using `	diagnostics_update_in_insert = false` fixed the issue.

PR #209, opened 2021-09-01
Only update diagnostic in Normal mode

ray-x:


lua/bufferline/diagnostics.lua
```diff
@@ -63,6 +66,7 @@ function M.get(opts)
       errors = d.errors,
     }
   end
+  last_diagnostics_result = result
```
akinsho, line 69: @ray-x this looks like an accidental global variable, which can easily pollute a users' global namespace or worse still conflict with an existing user variable and cause very weird explosions down the line

lua/bufferline/diagnostics.lua
```diff
@@ -53,6 +53,9 @@ function M.get(opts)
   if is_disabled(opts.diagnostics) then
     return setmetatable({}, mt)
   end
+  if vim.fn.mode(0) == 'i' and not opts.diagnostics_update_in_insert then
```
akinsho, line 56: This can actually be replaced with `api.nvim_get_mode().mode` which doesn't require the roundtrip through viml

akinsho (commented):
@ray-x thanks for the PR 🚀 much appreciated. I've left a comment about an accidental global var I think should be avoided and a minor style comment about the mode function. Other than those this looks good to me 👍🏿

lua/bufferline/diagnostics.lua
```diff
@@ -48,12 +50,22 @@ local function is_disabled(diagnostics)
   return not diagnostics or diagnostics ~= "nvim_lsp" or not vim.lsp.diagnostic.get_all
 end
 
+local function is_insert() -- insert or replace
+  local mode = vim.api.nvim_get_mode().mode
+  if mode == 'i' or mode == 'ic' or mode == 'ix'
```
akinsho, line 55: Why not just return the full conditional rather than use `if/else` e.g.
```lua
return mode == 'i' or mode == 'ic' or mode == 'ix'
```

lua/bufferline/diagnostics.lua
```diff
@@ -48,11 +50,20 @@ local function is_disabled(diagnostics)
   return not diagnostics or diagnostics ~= "nvim_lsp" or not vim.lsp.diagnostic.get_all
 end
 
+local function is_insert() -- insert or replace
+  local mode = vim.api.nvim_get_mode().mode
+  return mode == 'i' or mode == 'ic' or mode == 'ix'
```
akinsho, line 55: sorry to be a pain I actually just thought that this could be
```lua
return vim.tbl_contains(mode, {"i", "ic", "ix", "R", "Rc", "Rx"})
```

which is almost certainly marginally less performant but more readable.
Don't worry it's not a blocking issue just a thought

akinsho:
fixes #160

akinsho:
Merged, thanks again @ray-x 🚀

PR #212, opened 2021-09-02
fix(docs): correct name for buffer id on numbers section

NTBBloodbath:
Hey, hope you're doing well!

It took a few days for me to update by setup to match the recent "breaking changes" and I noticed that there was an inconsistency in the Numbers section of the docs while updating my configs so this PR aims to fix it.

Regards


NTBBloodbath:
Sure thing, didn't noticed before, thank you for pointing it!

PR #216, opened 2021-09-08
feat(diagnostics): support coc as source

xiyaowong:


xiyaowong:
I forgot to create a new branch...

akinsho:
@xiyaowong this looks good to me, I think it should be fine without a branch since I will squash and merge anyway. I'll test to make sure things aren't broken but will leave the verification of the `coc` logic to you since I don't use coc and don't really want to set it up just for this test case

xiyaowong:
> @xiyaowong this looks good to me, I think it should be fine without a branch since I will squash and merge anyway. I'll test to make sure things aren't broken but will leave the verification of the `coc` logic to you since I don't use coc and don't really want to set it up just for this test case

The current logic is basically fine, because similar PR has been merged to nvim-tree before, which is stable at present

akinsho (changes requested):
@xiyaowong just tried it out locally and got a crash using `nvim-lsp` due to a typo in the get all function name.

xiyaowong:
@akinsho sorry for my carelessness. fixed

akinsho:
Merged 👍🏿 thanks for contribution 🥇

PR #221, opened 2021-09-11
Extract Buffer Functions

LandonSchropp:
This is the first PR, primarily based around extracting the buffer logic from the `bufferline` function. Hopefully, this will pave the way for an easier transition to a `Tabpage` model.

I wanted to get early feedback on the direction of this work before polishing it off, so I opened this as a draft PR. Also, I’m still getting used to a Lua environment and development, so please let me know if I’m making any obvious mistakes with the language.

The main focus of this PR is a new `get_valid_buffers` function. This should return complete instances of `TabView` that are ready to be rendered. In the future, I plan on adding a `get_valid_tabpages` function that does the same thing, but with tabpages.

* How did you set up your environment for local Lua plugin development. I’m using LunarVim, which uses Packer, and I just added my local directory to the plugins table like this: `{ "~/Development/open-source/bufferline.nvim" }`. I’m not 100% sure that Packer is actually picking that up though. Did you do something similar, or is there a better way to handle this?
* How do I run the specs? I didn’t see any documentation on the framework used, but maybe I missed it somewhere.
* Should `tabpage` become a property of `TabView` rather than being an extra table parameter passed into `render`? This property obviously wouldn’t apply to `Tabpage`s, but it would apply to `Buffer` instances. This would simplify the interface for `render`, making it easier to include `Tabpage`s.
* What would you like to see done in terms of testing for this functionality. I tried not to change anything about how it works (mainly just extracting and moving bits around.
* How much needs to be changed in `render_buffer`? Would this function as a generic `render_tab_view` function as-is, or is it too specific to buffers?

### DIFF
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -103,50 +103,6 @@ function M.restore_positions()
   end
 end
 
---- sorts buf_names in place, but doesn't add/remove any values
---- @param buf_nums number[]
---- @param sorted number[]
---- @return number[]
-local function get_updated_buffers(buf_nums, sorted)
-  if not sorted then
-    return buf_nums
-  end
-  local nums = { unpack(buf_nums) }
-  local reverse_lookup_sorted = utils.tbl_reverse_lookup(sorted)
-
-  --- a comparator that sorts buffers by their position in sorted
-  local sort_by_sorted = function(buf_id_1, buf_id_2)
-    local buf_1_rank = reverse_lookup_sorted[buf_id_1]
-    local buf_2_rank = reverse_lookup_sorted[buf_id_2]
-    if not buf_1_rank then
-      return false
-    end
-    if not buf_2_rank then
-      return true
-    end
-    return buf_1_rank < buf_2_rank
-  end
-  table.sort(nums, sort_by_sorted)
-  return nums
-end
-
----Filter the buffers to show based on the user callback passed in
----@param buf_nums integer[]
----@param callback fun(buf: integer, bufs: integer[]): boolean
----@return integer[]
-local function apply_buffer_filter(buf_nums, callback)
-  if type(callback) ~= "function" then
-    return buf_nums
-  end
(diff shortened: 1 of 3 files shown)


akinsho:
@LandonSchropp should be free to answer in detail later today but, a few quick answers
* I use `plenary.nvim` to run the tests, I haven't actually documented it because no one else has really contributed continuously over time, so haven't gotten to documenting it yet.
* I personally think render buffer can't be made generic because it's too specific to buffers, I think a lot of commonalities could be re-used in one for tabs but I think making it generic will cause more problems than it's worth.
* Tbh I think the Tab model should be as similar as possible to the buffer but use the active window's buffer as it's "buffer" so when the filename prop is queried it should return the active buffers details etc. Not sure tabpage needs to be a top level prop definitely not on tabview which is super abstract since it can be used to render stuff that is not really related to tabs or buffers like group separators

akinsho:
Thinking about it a little I think that ultimately tabpages could/should end up looking a lot like the buffer model with the main differences being that most of it's properties are just proxies for the current active window's buffer.

```lua
---@class TabPageView
---@field filename string active buffers name
---@field path string active buffer's full path
---@field icon string active buffer's filetype icon
---@field tabpagenum number
---@field tabpage_windows number[] list of containings window
```
Just an example above but it could be quite similar and would make it much easier to align with the buffer stuff

 If you did it that way then you'd probably be able to re-use render buffer but some things would need to change, a lot of the settings are named around buffers, then handlers for click events are the buffer ones etc. not sure how to handle those. The settings can't be renamed since that would be a huge breaking change for existing users

LandonSchropp:
I ended up getting pulled into other projects, and unfortunately, I don't think I'm going to have time to finish this. I've since quit swimming upstream and switched to using buffers. (I remapped `:q` and `:wq`, which were my main annoyances.) Sorry for the trouble!

Issue #220, opened 2021-09-11
[Bug]: Error on VimEnter when using custom highlights


booperlv:
### What happened?

When opening neovim, I am greeted by an error message:
```
Error detected while processing VimEnter Autocommands for "*":
E5108: Error executing lua vim/shared.lua:219: after the second argument
: expected table, got string
stack traceback:
        vim/shared.lua:219: in function 'tbl_extend'
        vim/shared.lua:223: in function 'tbl_extend'
        vim/shared.lua:223: in function 'tbl_deep_extend'
        ...k/packer/start/bufferline.nvim/lua/bufferline/config.lua:90:
in function 'merge'
        ...k/packer/start/bufferline.nvim/lua/bufferline/config.lua:553:
 in function 'apply'
        ...ite/pack/packer/start/bufferline.nvim/lua/bufferline.lua:1202
: in function '__load'
        [string ":lua"]:1: in main chunk
```
Following the lines, it looked like it was an issue with custom highlights.
And so,, disabling/commenting my custom highlights out of my config worked.

### What did you expect to happen?

For it to launch properly, without any errors.

### Additional Information

I have not found any specific value in highlights that causes this, I haven't tried commenting out every single one yet as that seems to be unlikely to be the cause -  but what I know is just from a quick scroll in the codebase so ¯\\\_(ツ)_/¯

Here's my config for bufferline though (pardon for the mess, haven't cleaned the file up in a while)
```lua
require "bufferline".setup {
  options = {
    offsets = {{filetype = "NvimTree", text = ""}},
    buffer_close_icon = "",
    modified_icon = "",
    close_icon = " ",
    left_trunc_marker = " ",
    right_trunc_marker = " ",
    show_buffer_icons = true,
    show_buffer_close_icons = false,
    view = "multiwindow",
    separator_style = {"",""},
    indicator_icon = "",
  },

  highlights = {
    fill = {
      guifg = {attribute="fg",highlight="TabLine"},
      guibg = {attribute="bg",highlight="TabLine"}
    },
    background = {
      guifg = {attribute="fg",highlight="TabLine"},
      guibg = {attribute="bg",highlight="TabLine"}
    },

    buffer_selected = {
      guifg = {attribute="fg",highlight="TabLineSel"},
      guibg = {attribute="bg",highlight="TabLineSel"},
      gui = "none"
    },
    buffer_visible = {
      guifg = {attribute="fg",highlight="TabLine"},
[...]

PR #222, opened 2021-09-12
fix: convert user config highlights before merging

akinsho:
@booperlv can you please try this branch and see if it fixes your issue in #220

### DIFF
--- a/lua/bufferline/config.lua
+++ b/lua/bufferline/config.lua
@@ -2,6 +2,16 @@ local M = {}
 
 local fmt = string.format
 
+---@class DebugOpts
+---@field logging boolean
+
+---@class GroupOptions
+---@field toggle_hidden_on_enter boolean re-open hidden groups on bufenter
+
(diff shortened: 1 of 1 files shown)


booperlv:
Apologies for the late reply, it did work! For whatever reason though, it still does not update the colors of the bufferline when changing colors using custom highlights (used to have a `dofile` on the ColorScheme autocommand to solve this). It does solve the issue at hand though!

akinsho:
@booperlv can you have another go with it? I think it should work correctly now.

akinsho:
fixes #220

akinsho:
Going to merge this since it's a bit of a blocker for anyone on 0.5 using the custom highlights, which isn't ideal. Let me know if you test this later and still have issues.

booperlv:
Apologies for not being able to reply, was sleeping during that time. I'll update or create a new issue if I find anything at the end of the day! Thanks!

PR #234, opened 2021-09-28
fix(offset): Search winhl in reverse order when guessing hl.

sindrets:
When `:h 'winhl'` contains multiple entries for the same hl-group, vim will
use the last one. Thus, we should search the winhl entries in reverse
order when we are trying to guess the offset hl.

Example:
- The value of `'winhl'` is `Normal:Foo,Normal:Bar,Normal:Baz`.
- Vim will in this case link `Normal` to `Baz`, but the current implementation will guess `Foo`.


akinsho:
Thanks for the PR @sindrets and for catching this issue 👍🏿 LGTM

PR #236, opened 2021-09-30
feat: split buffer with pick

lu5je0:
Add a command to split buffer by pick

https://user-images.githubusercontent.com/17700877/135425614-7c92525c-0587-46fc-983e-8524dcedb214.mov

.

### DIFF
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -228,6 +228,12 @@ local function select_buffer_apply(func)
   refresh()
 end
 
+function M.split_buffer_with_pick()
+  select_buffer_apply(function(buf_id)
+    vim.cmd("vertical sb" .. buf_id)
+  end)
+end
+
 function M.pick_buffer()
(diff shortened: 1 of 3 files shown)


akinsho:
Hi @lu5je0 thanks for the contribution, tbh I was on the fence about whether to add this primarily because I don't want to try and add all the possible mechanisms e.g. `split`, `tabedit` etc. to the plugin itself if a user can just do what you did here by themselves in their config. Since another user might come by later and say I want a function for horizontal splits or I want one to open a new tab etc 🤷🏿‍♂️ .

This looks good to me but I think it could also just be part of the documentation so a user can easily figure out how to do things like this themselves without it having to be added to the plugin itself. What do you think?

akinsho:
@lu5je0 are you still working on this PR? or just using this as your fork?

lu5je0:
Sorry, I am just using this as fork for now, so I'm gonna close this.

PR #257, opened 2021-11-14
feat: Add hint diagnostic highlight

toppair:
Looks like hint level diagnostics are not explicitly set to derive hl from corresponding existing groups, so I did that. Not sure about the fallback, seems suitable though.

### DIFF
--- a/lua/bufferline/config.lua
+++ b/lua/bufferline/config.lua
@@ -223,6 +223,7 @@ local function derive_colors()
   local error_hl = nightly and "DiagnosticError" or "LspDiagnosticsDefaultError"
   local warning_hl = nightly and "DiagnosticWarn" or "LspDiagnosticsDefaultWarning"
   local info_hl = nightly and "DiagnosticInfo" or "LspDiagnosticsDefaultInformation"
+  local hint_hl = nightly and "DiagnosticHint" or "LspDiagnosticsDefaultHint"
 
   local error_fg = hex({
     name = error_hl,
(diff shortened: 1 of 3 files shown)


lua/bufferline/config.lua
```diff
@@ -242,6 +243,12 @@ local function derive_colors()
     fallback = { name = "Normal", attribute = "fg" },
   })
 
+  local hint_fg = hex({
+    name = hint_hl,
+    attribute = "fg",
+    fallback = { name = "SpecialComment", attribute = "fg" },
```
akinsho, line 249: I don't believe `SpecialComment` is a generic enough highlight as checking my nvim setup I don't actually have this defined, and it isn't one of the base ones that nvim defines by default. The fallback should be something that nvim defines by default, although that doesn't guarantee a user's colorscheme has highlighted it, but it should have. I think maybe `Directory` is a good enough fallback. It won't be great for all but tbh a fallback isn't really necessary for a lot of these any more since the Lsp highlights are much better covered nowadays than when I first released the plugin

akinsho (changes requested):
Thanks for the PR @toppair this looks good to me except the fallback highlight which I don't think is general enough. If you change it to a highlight defined by default, then this will be good to merge. I've suggested trying `Directory` the choice is a bit arbitrary but shouldn't really matter since the fallback is unlikely to be needed unless someone is using a custom or outdated colorscheme

lua/bufferline/config.lua
toppair: Yes, Directory is a better choice then. Applied it.

akinsho:
Awesome thanks for making the changes and for the contribution 👍🏿

PR #272, opened 2021-12-09
fix: use vim.diagnostic.get instead of vim.lsp.diagnostic.get_all

### LINKED ISSUE #271: Replace deprecated `vim.lsp.diagnostic.get_all` call(s)
https://github.com/akinsho/bufferline.nvim/blob/463637a3ac86dcaacbcd47aa608f53aaad749696/lua/bufferline/diagnostics.lua#L54

Is using `vim.lsp.diagnostic.get_all`, which has been deprecated. `vim.diagnostic.get()` should be used instead.

hood:
Closes #271.

### DIFF
--- a/lua/bufferline/diagnostics.lua
+++ b/lua/bufferline/diagnostics.lua
@@ -51,7 +51,7 @@ local function is_disabled(diagnostics)
   if
     not diagnostics
     or not vim.tbl_contains({ "nvim_lsp", "coc" }, diagnostics)
-    or (diagnostics == "nvim_lsp" and not vim.lsp.diagnostic.get_all)
+    or (diagnostics == "nvim_lsp" and not vim.diagnostic.get)
     or (diagnostics == "coc" and vim.g.coc_service_initialized ~= 1)
   then
     return true
@@ -66,7 +66,7 @@ end
 
 local get_diagnostics = {
   nvim_lsp = function()
-    return vim.lsp.diagnostic.get_all()
+    return vim.diagnostic.get()
   end,
 
   coc = (function()
--- a/lua/bufferline/diagnostics.lua
+++ b/lua/bufferline/diagnostics.lua
@@ -1,3 +1,5 @@
+local utils = require("bufferline.utils")
+
 local M = {}
 
 local fn = vim.fn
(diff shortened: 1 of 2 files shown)


akinsho:
@hood thanks for picking this up and for raising the PR. Is this problem exclusive to nightly nvim or does it affect nvim 0.6 since I'm running 0.6 and don't see any issues using the current `lsp` api but using the `diagnostic.get` I don't see any diagnostic indicators in buffer tabs any more i.e. this PR seems to break this functionality in 0.6 stable

hood:
Good point. The deprecation has taken place at 0.7, in one of the latest commits. Could we maybe get this merged in a dedicated branch, for users who like to be on the latest nightly versions?

hood:
Now that I think about it, we could maybe simply add a simple check whether the user's NeoVim version supports vim.diagnostic  or vim.lsp.diagnostic. I'll try to wrap my head around it asap, but I'm pretty busy at the moment, so anyone willing to give a try is welcome.

akinsho:
@hood the fix is actually probably quite straightforward. You can check for a nightly nvim by running `vim.fn.has('nvim-0.7')` and use the diagnostic method if nightly otherwise use the current mechanism. Shouldn't need to be more complex than that.

hood:
the check has been implemented

lua/bufferline/diagnostics.lua
```diff
@@ -66,7 +67,11 @@ end
 
 local get_diagnostics = {
   nvim_lsp = function()
-    return vim.lsp.diagnostic.get_all()
+    if vim.fn.has('nvim-0.7') then
```
akinsho, line 70: The same applies here as well

akinsho (changes requested):
@hood thanks for making the tweaks. I've left a comment about a bug in the changes that would still cause issues for anyone on the latest stable build

lua/bufferline/diagnostics.lua
hood: Sorry for not testing thoroughly, unfortunately I've got a very limited bandwith to work on this issue. I've added a `is_truthy` utility function and I've used it to wrap `vim.fn.has` usages.

akinsho:
@hood thanks for tweaking this. I just tried it and it still had a few issues so I pushed a fix to your branch but doesn't seem to be reflected here. Anyway the diff of my changes is

```diff
diff --git a/lua/bufferline/diagnostics.lua b/lua/bufferline/diagnostics.lua
index 438df7b..c322cb8 100644
--- a/lua/bufferline/diagnostics.lua
+++ b/lua/bufferline/diagnostics.lua
@@ -49,12 +49,15 @@ local mt = {
   end,
 }
 
+local is_nightly = utils.is_truthy(fn.has("nvim-0.7"))
+local is_valid_version = utils.is_truthy(fn.has("nvim-0.5"))
+
 local function is_disabled(diagnostics)
   if
     not diagnostics
     or not vim.tbl_contains({ "nvim_lsp", "coc" }, diagnostics)
-    or (diagnostics == "nvim_lsp" and utils.is_truthy(vim.fn.has('nvim-0.7')) and not vim.diagnostic.get) -- <-- Anything 0.7+
-    or (diagnostics == "nvim_lsp" and not utils.is_truthy(vim.fn.has('nvim-0.7')) and not vim.lsp.diagnostic.get) -- <- Anything below 0.7
+    -- check if the current nvim version is one that will have either vim.diagnostics or vim.lsp.diagnostics
+    or (diagnostics == "nvim_lsp" and not is_valid_version)
     or (diagnostics == "coc" and vim.g.coc_service_initialized ~= 1)
   then
     return true
@@ -69,11 +72,11 @@ end
 
 local get_diagnostics = {
   nvim_lsp = function()
-    if utils.is_truthy(vim.fn.has('nvim-0.7')) then
+    if is_nightly then
       return vim.diagnostic.get()
-    else
-      return vim.lsp.diagnostic.get()
     end
+    ---@diagnostic disable-next-line: deprecated
+    return vim.lsp.diagnostic.get_all()
   end,
 
   coc = (function()
```
Can you apply the changes locally and if it works for you then can you commit them and I'll merge the branch.

PR #276, opened 2021-12-15
Fix diagnostic buffer number for nightly nvim

c0r73x:


lua/bufferline/diagnostics.lua
```diff
@@ -73,7 +73,18 @@ end
 local get_diagnostics = {
   nvim_lsp = function()
     if is_nightly then
-      return vim.diagnostic.get()
+      results = {}
```
akinsho, line 76: This should be
```suggestion
      local results = {}
````
Otherwise the `results` variable will be a global variable and almost certainly clash with something

Comment (c0r73x) [reply lua/bufferline/diagnostics.lua]: Fixed now, sorry about that.. I'm used to writing moonscript :P

File: lua/bufferline/diagnostics.lua [hunk 2]
```diff
@@ -73,7 +73,18 @@ end
 local get_diagnostics = {
   nvim_lsp = function()
     if is_nightly then
-      return vim.diagnostic.get()
+      local results = {}
+      diagnostics = vim.diagnostic.get()
```
akinsho, line 77: The same applies here as well actually, I only just saw this one
```suggestion
      local diagnostics = vim.diagnostic.get()
```

akinsho (approved):
@c0r73x I haven't tested this because I don't use the nightly builds so I can't verify that this actually works but the code looks fine to me 👍🏿 🤞🏿 thanks.

PR #284, opened 2021-12-31
chore: no extra padding if numbers func returns empty string

horseinthesky:
Small buffer index padding fix - if `numbers` is a function that returns an empty string we don't need additional padding.

I am using it in the following example 
```
numbers = function(opts)                                         
  if vim.api.nvim_get_current_buf() == opts.id then return "" end
  return opts.ordinal                                            
end,                                                             
```

Before - extra padding
image

After - no extra padding
image


akinsho (approved):
LGTM 👍🏿 thanks @horseinthesky

Issue #297, opened 2022-01-09
[Bug]: Filtering as covered in README shouldn't exists


lpawlak1:
### What happened?

The issue is that based on code parameters mismatch

```lua
---@param callback fun(buf: integer, bufs: integer[]): boolean
```
https://github.com/akinsho/bufferline.nvim/blob/master/lua/bufferline.lua#L135

Even though this takes bufs as parameter in README function provided doesn't take second argument.

### What did you expect to happen?

Filter buffers

### Additional Information

I would suggest changing README adding information about filtering function and/or changing code to match function from README

akinsho:
Closing this as the associated PR is merged

PR #298, opened 2022-01-09
docs: add buf_numbers argument to README

lpawlak1:
https://github.com/akinsho/bufferline.nvim/issues/297

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,7 +246,7 @@ require('bufferline').setup {
         return true
       end
       -- filter out by it's index number in list (don't show first buffer)
-      if buf_numbers[1] ~= buf_numbers then
+      if buf_numbers[1] ~= buf_number then
         return true
       end
     end,
--- a/README.md
+++ b/README.md
@@ -231,7 +231,7 @@ require('bufferline').setup {
       return "("..count..")"
     end,
     -- NOTE: this will be called a lot so don't do any heavy processing here
-    custom_filter = function(buf_number)
+    custom_filter = function(buf_number, buf_numbers)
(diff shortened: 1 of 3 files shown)


akinsho (commented):
@lpawlak1 thanks for the PR but I don't think it's necessary. The buffer numbers are actually passed in on purpose and are depended on by users, including myself. They don't need to be removed.  Is there a reason why you're removing them exactly?

lpawlak1:
Based on these lines (README.md)
```lua
-- NOTE: this will be called a lot so don't do any heavy processing here
    custom_filter = function(buf_number)
      -- filter out filetypes you don't want to see
      if vim.bo[buf_number].filetype ~= "<i-dont-want-to-see-this>" then
        return true
      end
      -- filter out by buffer name
      if vim.fn.bufname(buf_number) ~= "<buffer-name-I-dont-want>" then
        return true
      end
      -- filter out based on arbitrary rules
      -- e.g. filter out vim wiki buffer from tabline in your work repo
      if vim.fn.getcwd() == "<work-repo>" and vim.bo[buf_number].filetype ~= "wiki" then
        return true
      end
    end,
```

it was unclear to my why to use second parameter (list) in filtering function, but if irl it is used it should stay and probably second paramer should be used in example configuration as its unclear that it actually can be used

lpawlak1:
If something as simple as added 
```lua
-- filter out by it's index number in list (don't show first buffer)
      if buf_numbers[1] ~= buf_number then
        return true
      end
```
can be useful to know that this exists (without looking into code)

PR #301, opened 2022-01-10
trim unnecessary (too much) front-padding in slant style buffer "tab"

errantepiphany:
When the buffer "tab" style is slant, the front of the "tab" has one extra space too much. It is unnecessary and just takes up "space" (pun intended).

Signed-off-by: David Ward <dward@redhat.com>


errantepiphany:
This isn't right... It causes a problem bouncing back and forth when selected or not. I'll fix and resubmit...

PR #303, opened 2022-01-12
fix(README): replace deprecated api

cathaysia:


akinsho:
Thanks a lot @cathaysia

PR #308, opened 2022-01-15
docs: fix inconsistent highlight group quoting

mcauley-penney:
- highlight groups, under the "HIGHLIGHTS"
      section, had inconsistent quoting, e.g.
      guifg = <color-value-here> versus
      guifg = '<color-value-here>'. All
      quoting in example highlight groups should
      now be fixed


akinsho:
Thanks 👍🏿

PR #321, opened 2022-02-05
feat(groups.lua): add "gui" option to group label

mcauley-penney:
- allows for the user to specify a gui option for a group label on a per-group basis, e.g. "bold" to change name of group to bold text
    - defaults to "none" to preserve originally intended UI appearance


mcauley-penney:
I just realized that this is actually described in the readme, in the `groups` section, but I couldn't manage to make it work without this modification.

PR #322, opened 2022-02-07
refactor: explicitly require the bit module

drscream:
This makes the dependency explicit, limits the scope to the current file and provides faster access to the bit.* functions, too. It's good programming practice not to rely on the global variable bit being set (assuming some other part of your application has already loaded the module). The require function ensures the module is only loaded once, in any case.

Without explicit adding the line I receive:

```
lua/bufferline/colors.lua:71: attempt to index global 'bit' (a nil value)
```

### DIFF
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -1,4 +1,3 @@
-local bit = require("bit")
 local constants = require("bufferline.constants")
 local utils = require("bufferline.utils")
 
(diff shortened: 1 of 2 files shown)


akinsho:
@drscream thanks for the PR, so the `bit` module is expected to be available when using the specific version of Lua that nvim ships with. This unfortunately has ended up being a recurrent issue because some distros of Linux or different OSes ship with slightly different versions, and so modules which a plugin author should be able to depend on to exist globally sometimes don't. Depending on `bit` isn't dissimilar to depending on the `math` object. If it reduces issues and doesn't break anything, then LGTM to add

PR #323, opened 2022-02-13
perf(coc-diagnostics): use event `CocDiagnosticChange`

xiyaowong:
Coc will trigger `CocDiagnosticChange` when diagnostic status changed.
Only get diagnostics when `CocDiagnosticChange`, can avoid a lot of
unnecessary `CocActionAsync` calls.

### DIFF
--- a/lua/bufferline/diagnostics.lua
+++ b/lua/bufferline/diagnostics.lua
@@ -93,7 +93,7 @@ local get_diagnostics = {
   coc = (function()
     local diagnostics = {}
 
-    _G.___refresh_coc_diagnostics_for_bufferline___ = function()
+    function M.refresh_coc_diagnostics()
       pcall(fn.CocActionAsync, "diagnosticList", function(err, res)
         if err ~= vim.NIL then
           return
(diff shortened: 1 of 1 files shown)


xiyaowong:
Although function and event are defined whether or not coc is used, I don't see a big impact:smile:

lua/bufferline/diagnostics.lua
```diff
@@ -92,31 +92,34 @@ local get_diagnostics = {
...
+          if bufnr ~= -1 then
+            result[bufnr] = result[bufnr] or {}
+            table.insert(result[bufnr], { severity = diagnostic.level })
+          end
+        end
+        diagnostics = result
+      end)
     end
+
+    vim.cmd([[autocmd User CocDiagnosticChange lua ___refresh_coc_diagnostics_for_bufferline___()]])
```
akinsho, line 120: Rather than create a global variable I'd prefer if this just called the function inside bufferline e.g.
```lua
vim.cmd([[autocmd User CocDiagnosticChange lua require('bufferline.diagnostics').refresh_coc_diagnostics()]])
```

akinsho:
@xiyaowong thanks for the PR 👍🏿 just made one comment about converting the function to be an export of the module and use that in the autocommand rather than a global variable

xiyaowong:
maked a change @akinsho

akinsho (approved):
LGTM @xiyaowong will merge this in, presuming you've tested it, and it all works. I don't use `coc` so can't test it out myself

PR #333, opened 2022-03-08
fix(offsets): use nvim_eval_statusline to calculate the size

xiyaowong:


xiyaowong:
`strwidth` returns width of the raw text which includes expression, highlight. It's inaccurate

akinsho:
Ah that makes sense @xiyaowong I hadn't expected anyone to put raw tabline components into that space

akinsho:
@xiyaowong just to clarify have you also seen the that you can pass highlights to the custom area without adding them to the text e.g.
```lua
custom_areas = {
	return {{text = "this is text", guifg =  "#EC5241"}}
}
```

xiyaowong:
yes, I known it

xiyaowong:
image
This is my custom area

akinsho:
@xiyaowong that API wasn't available in the stable nvim version, just FYI it would have broken for anyone using stable nvim

Issue #353, opened 2022-04-01
[Bug]: Tab mode is not working because of missing visibility method.


buraksirma:
### What happened?

Bufferline is failing in tabooed because of missing method.

### What did you expect to happen?

Render tabpages.

### Additional Information

...

PR #354, opened 2022-04-01
fix: missing visibility method on TabPage

buraksirma:
fix: missing visibility method on TabPage

Added missing method for TabPages.

See: #353


akinsho (approved):
LGTM thanks for the fix

PR #369, opened 2022-04-10
CI: update github workflow (use main branch)

justin-f-perez:
I'm unsure but I would guess the primary branch for bufferline changed very recently- it happened to break my nvim config when I updated today (because of how packer handles branches) and as I was poking around trying to figure out what happened I noticed these two files still referencing master.

(PS: thanks for the great plugin!)

### DIFF
--- a/.github/workflows/formatting.yaml
+++ b/.github/workflows/formatting.yaml
@@ -1,7 +1,7 @@
 name: stylua
 on:
   push:
-    branches: ["master"]
+    branches: ["main"]
     paths-ignore:
       - ".github/**"
       - "*.md"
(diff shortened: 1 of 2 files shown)


akinsho:
Thanks for catching this @justin-f-perez, tbh I hadn't expected this to break for people, I kind of thought packer and other package managers would just start pulling from `main` since that was the new default but tbh I didn't know for sure

justin-f-perez:
@akinsho I thought so too! IMO this is a bug in Packer- I was surprised it hard-codes a default. I suggested to packer maintainer they should check `ls-remote HEAD` instead

PR #372, opened 2022-04-10
fix: always_show_bufferline should work in tabs mode

### LINKED ISSUE #371: Bufferline is not shown when mode == "tabs" and always_show_bufferline == false
When in `tabs` mode and flag `always_show_bufferline` is set to false, bufferline (tabline) is not shown because the check relies on number of buffers open, instead of relying on the number of tabs open:
https://github.com/akinsho/bufferline.nvim/blob/50e1bfe6f2c474c0a6e8171606b001f3b17ddeb2/lua/bufferline.lua#L123

Expected behavior:
Tabline shows when number of tabs is greater than 1, and hidden otherwise.

agorgl:
Fixes #371

### DIFF
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -118,9 +118,10 @@ local function bufferline()
   return tabline
 end
 
---- If the buffer count has changed and the next tabline status is different then update it
+--- If the item count has changed and the next tabline status is different then update it
 function M.toggle_bufferline()
-  local status = (config.options.always_show_bufferline or utils.get_buf_count() > 1) and 2 or 0
+  local item_count = config:is_tabline() and utils.get_tab_count() or utils.get_buf_count()
+  local status = (config.options.always_show_bufferline or item_count > 1) and 2 or 0
   if vim.o.showtabline ~= status then
     vim.o.showtabline = status
   end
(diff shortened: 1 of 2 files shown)


akinsho (changes requested):
@agorgl thanks for the PR this looks good to me I had a quick variable name change to suggest. I tried pushing it but can't add to this branch 
the diff is
```
diff --git a/lua/bufferline.lua b/lua/bufferline.lua
index 8f84179..6527f3c 100644
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -120,8 +120,8 @@ end
 
 --- If the item count has changed and the next tabline status is different then update it
 function M.toggle_bufferline()
-  local nitems = config.options.mode == "tabs" and utils.get_tab_count() or utils.get_buf_count()
-  local status = (config.options.always_show_bufferline or nitems > 1) and 2 or 0
+  local item_count = config:is_tabline() and utils.get_tab_count() or utils.get_buf_count()
+  local status = (config.options.always_show_bufferline or item_count > 1) and 2 or 0
   if vim.o.showtabline ~= status then
     vim.o.showtabline = status
   end
```
I have convenience method for checking if it's a tabline which I think it's better to use here

agorgl:
Fixed as suggested!

akinsho (changes requested):
@agorgl thanks for making the tweaks, the tags file should not be added to version control though

agorgl:
Ah sorry it was a blind git add, fixing

akinsho (approved):
LGTM 🚀 🥇 thanks

PR #373, opened 2022-04-11
fix: correctly re-export and document exec function

ExpandingMan:
There seems to have been a few changes to the API that did not get documented in the README.  Wasn't too hard to reverse engineer, so when I figured it out I changed the README to reflect what I found.

### DIFF
--- a/README.md
+++ b/README.md
@@ -675,14 +675,14 @@ A user can also execute arbitrary functions against a buffer using the
 `exec` function. For example
 
 ```lua
-    require('bufferline.commands').exec(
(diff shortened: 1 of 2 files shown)


akinsho:
@ExpandingMan thanks for the PR, it turns out I forgot to update the docs and also re-export the function from the commands' module when I refactored the plugin to add tabpages. I wanted to remove references to functions that specifically refer to buffers, since some/most of those will work for tabs eventually.

I've removed the `require` from the `commands` module since I generally don't want users to access private modules that hold implementation details

PR #403, opened 2022-05-01
fix(config): remove mappings field from defaults

### LINKED ISSUE #402: [Bug]: getting alert about `mappings` deprecation even when mappings are not defined
### What happened?

Every time I compile Packer using `:PackerCompile` I get a deprecation warning notification from Bufferline.

https://user-images.githubusercontent.com/551858/166161643-dd9b889e-e8ca-42e8-9542-eb8a2021a4a0.mp4

It appears that the `mappings = false` from the defaults is getting loaded in by Bufferline and that triggers the alert. 

Expected behavior
Since this is not an option I defined, the message should not show up.

### Additional Information

_No response_

### commit

d343bc322e00e9b21bcb9b3c8bf6a989b64a92a2

dkarter:
Fixes #402

### DIFF
--- a/lua/bufferline/config.lua
+++ b/lua/bufferline/config.lua
@@ -581,7 +581,6 @@ local function get_defaults()
       name_formatter = nil,
       tab_size = 18,
       max_name_length = 18,
-      mappings = false,
       color_icons = true,
       show_buffer_icons = true,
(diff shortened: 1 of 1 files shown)


akinsho:
@dkarter thanks for the PR the change looks good to me

PR #415, opened 2022-05-11
Cterm color support

girishji:
Please verify changes.


doc/bufferline.txt
```diff
@@ -823,6 +828,24 @@ highlight command. See `:h highlight` .
         };
     }
 <
+When `termguicolors` is not enabled, user is expected to specify color values
```
akinsho, line 831:
```suggestion
When `termguicolors` is not enabled, a user is expected to specify color values
```

doc/bufferline.txt
```diff
@@ -823,6 +828,24 @@ highlight command. See `:h highlight` .
...
+similar to the format above (use `ctermbg` and `ctermfg` instead). User can
+also directly specify [color
+numbers](https://www.ditig.com/256-colors-cheat-sheet).
+
+for example: >
+    highlights = {
+        fill = {
+            ctermbg = 7
+            ctermfg = 0
+            }
```
akinsho, line 841:
```suggestion
          }
```

doc/bufferline.txt
```diff
@@ -823,6 +828,24 @@ highlight command. See `:h highlight` .
         };
     }
 <
+When `termguicolors` is not enabled, user is expected to specify color values
+similar to the format above (use `ctermbg` and `ctermfg` instead). User can
```
akinsho, line 832:
```suggestion
similar to the format above (use `ctermbg` and `ctermfg` instead). A user can
```

lua/bufferline/ChangeLog
```diff
@@ -0,0 +1,3256 @@
+Author: Girish Palya <girishji@gmail.com>
```
akinsho, line 1: This seems like an unrelated addition? Is this intentional? If I were to introduce a changelog, I'd expect it to be part of a separate PR and discussed separately

lua/bufferline/config.lua
```diff
@@ -94,13 +94,22 @@ local function convert_highlights(map)
...
+          else
+            updated[hl][attribute] = colors.get_hex({
+              name = value.highlight,
+              attribute = value.attribute,
+            })
+          end
         else
-          updated[hl][attribute] = nil
-          utils.notify(fmt("removing %s as it is not formatted correctly", hl), utils.W)
+          if not string.find(attribute, "cterm") then
```
akinsho, line 109: Why is this change necessary, i.e. why does `cterm` require different error notification handling than hex here

lua/bufferline/ui.lua
```diff
@@ -257,7 +257,9 @@ local function highlight_icon(buffer, color_icons, hl_defs)
   local new_hl = highlights.generate_name(hl, { visibility = state })
   local guifg = not color_icons and "fg" or colors.get_hex({ name = hl, attribute = "fg" })
   local guibg = colors.get_hex({ name = bg_hls[state], attribute = "bg" })
-  highlights.set_one(new_hl, { guibg = guibg, guifg = guifg })
+  local ctermfg = colors.get_cterm_color({ name = bg_hls[state], attribute = "fg" })
```
akinsho, line 260: Can the behaviour of `cterm` be internalised into `get_hex` and you just use a different flag e.g.
```lua
get_color({..., type = 'cterm'})
```
Or something. I don't like that the implementation details of thinking about cterm vs gui has leaked into the rest of the code base. It will mean that going forward I have to think about specifically getting cterm colors or not.

I think a necessary criteria of this plugin is that it does not complicate future development for me since if the maintenance burden is high this will not work for me.

akinsho: Actually, maybe `get_color` should just always return the `cterm` value instead of special casing cterm with a separate function

akinsho (changes requested):
@girishji thanks for the PR. I will need to *thoroughly* test it because it *absolutely* must not break for other users not using `cterm`. One very important point is that I don't want to have choose between `get_cterm` color and `get_hex` in the future. This is too high a maintenance burden since I will have to now specifically reason about where to use it etc.

I think `get_hex` → `get_colour` and this should automatically return cterm values. Try to understand once this I is merged I will have to deal with this situation for maybe years, so I do not want to have to do a lot of if statements for cterm, because ultimately I will forget and probably not do it. Does this make sense

lua/bufferline/ChangeLog
girishji: Not intentional. It appeared as a consequence of a script I use for checkin.

lua/bufferline/config.lua
girishji: removed this check. i think i introduced it when i was playing with return value of -1 (didn't know about lua's nil). i was learning lua as i was playing with this.

doc/bufferline.txt
girishji: agree

doc/bufferline.txt
girishji: agree

doc/bufferline.txt
girishji: agree

lua/bufferline/ui.lua
girishji: you are using `name = hl` and not `bg_hls` for `guifg`. These things have different values that are not in end user's control. I would like to disassociate from this completely and have user specify it.

girishji: It is better to separate `get_hex` from `get_color` and deal with extra cognitive overhead for now. Other option is to encapsulate both into one object which emits a hex `string` or color number `integer`. This way you can use it consistently everywhere, including in color_spec.

girishji: you can review now, thx

girishji:
added changes. you can review, thx!

akinsho:
@girishji can you point this PR at `dev` instead of main so it can be tested thoroughly on that branch before going into `main`

ChangeLog
```diff
@@ -1,10 +1,21 @@
 Author: Girish Palya <girishji@gmail.com>
```
akinsho: This file is still present 😅

akinsho: Ignore me I just saw the next commit

akinsho (changes requested):
Added a few more small comments and one big one around the direction. I think that there shouldn't be two separate functions. I really do not want to have to deal with two different types of colour function. I think that cterm can just be an argument to get colour and the value return depend on that. I think you could create a wrapper type and type annotate it so it is clearer or you could return a `string | number`

girishji:
I will get to this today

lua/bufferline/ui.lua
```diff
@@ -257,7 +257,12 @@ local function highlight_icon(buffer, color_icons, hl_defs)
   local new_hl = highlights.generate_name(hl, { visibility = state })
   local guifg = not color_icons and "fg" or colors.get_hex({ name = hl, attribute = "fg" })
   local guibg = colors.get_hex({ name = bg_hls[state], attribute = "bg" })
-  highlights.set_one(new_hl, { guibg = guibg, guifg = guifg })
```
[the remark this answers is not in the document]
girishji: You have another parameter to `get_hex()` called `fallback`. How'd you account fo r that?

PR #424, opened 2022-05-18
docs(README): add a note for GUI users to the FAQ.

floriankraft:
For Windows users like me, who didn't know, that there is a distinct config file for disabling the GUI Tabline in Neovim. Took me a while to figure this one out. 😅

### DIFF
--- a/README.md
+++ b/README.md
@@ -744,6 +744,10 @@ not block rendering the tabline.
   If you are using `airline` make sure you set `let g:airline#extensions#tabline#enabled = 0`.
   If you are using `lightline` this also takes over the tabline by default and needs to be deactivated.
 
+  If you are on Windows and use the GUI version of nvim (nvim-qt.exe) then also ensure, that `GuiTabline`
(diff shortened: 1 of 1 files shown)


akinsho:
@floriankraft I'm not sure if I understand what this. It sounds like this change is about using the GUI in windows? Rather than using the terminal? Since the `GuiTabline` is only relevant if using some external GUI. In which case, I don't think this advice is generally applicable and is only relevant to anyone using a GUI but still wanting to use this plugin which is primarily aimed at TUI given that GUIs tend to do their own fancy tabline stuff.

akinsho:
Ah apologies, I just read your comment in more detail and it's fine You do clarify about using a GUI

floriankraft:
Thanks man! 😄 I was about to write a lengthy text explaining my motivations etc. etc. 😁 

I think people like me who are (sometimes) forced to use the GUI would not miss the default GUI tabbar when they are used to the TUI looks (and have a plugin like yours installed).

Issue #425, opened 2022-05-18
[Feature]: Nice bufferline-styled `winbar`


mrjones2014:
This morning, a PR merged to Neovim that allows us to set a per-window toolbar (a `winbar`): https://github.com/neovim/neovim/pull/18562

It would be awesome to have Bufferline handle setting the `winbar` window option, to the currently opened buffer name in each window, with the snazzy bufferline styles.

Here's a quick example I did, but note that I didn't add any nice styling that I'd want Bufferline to handle.

![image](https://user-images.githubusercontent.com/8648891/169031733-2c26204a-d568-4a4c-b9a9-8fd76847945c.png)

mrjones2014:
Feel free to close if you feel this is out of scope for bufferline.nvim or that it should just be it's own separate plugin :)

mrjones2014:
Alternatively, could we just show the entire bufferline in the `winbar` area, and have each `winbar` show the active buffer like Bufferline does, but have the active buffer indicator be window-specific?

mrjones2014:
It seems to "just work" quite nicely if I manually run this command in each window: `:lua vim.wo.winbar=_G.nvim_bufferline()` then run `:set showtabline=0`. It gets the active buffer in each window correct, right out of the box.

So maybe what we should do is just have an option to display the entire bufferline.nvim into the `winbar` of each window, rather than the tabline?

(note, I changed my style config a little in between screenshots)
![image](https://user-images.githubusercontent.com/8648891/169039245-3dd3f4aa-5663-4598-8e97-d676cbed967d.png)

mrjones2014:
I personally would love to have bufferline.nvim render per-window (basically _instead of_ incline.nvim), but if you feel it's out of scope or not useful, that's fine too 👍

akinsho:
@mrjones2014 I'm happy to review/help guide a PR if you are willing to look into it, I guess what I'm envisioning if we go down this route it two modes i.e.
a `window` mode which is no global bufferline just winbars that show the one buffer in the bufferline style in each window.

The other mode would be a mix of tabline + winbars, but if a window already has a winbar then it's buffer shouldn't be included in the tabline since that's redundant.

I don't mind this addition if it's done in a way that is completely opt in and not too fancy.

ATM in `dev` you can pass a list of buffers to the `ui.tabline` function and that will create a tabline string (it could be worth passing it options to skip some steps such as custom areas etc. not sure if that's needed. I can see the implementation looking something like

```lua
[...]

Issue #348, opened 2022-03-29
[Feature Request]: Show buffer line in all (vertical) splits?


mrjones2014:
Is something like this edited image possible? To be clear, I took 2 separate screenshots and edited them together to get this image:

![CleanShot-2022-03-29-at-13 04 56](https://user-images.githubusercontent.com/8648891/160666863-e404368a-bab4-42b7-ab10-438449357240.png)

Note that it shows the active buffer for each window, separately.

Not sure if this is possible or feasible but I'd love to see it.

akinsho:
@mrjones2014 so this is a duplicate of a very recent feature request #345. I first saw this behaviour in Emacs' centaur tabs, which originally inspired this plugin. I've personally since come off of it for a few reasons.

Firstly it's useful in `vscode` or `emacs` because both of those have the concept of `editor groups` so rather than windows it's the files are actually specifically related to each section, whereas in vim it would literally be 20 buffers open for example and 2 splits, what should happen in that case how do all the other buffers arrange themselves?
How does it behave if there is a horizontal split on top of two vertical ones?

Even if they could be arranged sensibly, why would that be useful?
At this point for nvim it seems more like eye candy than anything that would actually improve the editing experience.

Happy to hear a detailed explanation on how it improves the editing experience, though, since I'd only personally consider this being added if it was something that a significant amount of bufferline users might want to use 🤷🏿‍♂️ or at least I could explain the utility to someone else.

Re. implementation, I guess it would look something like the logic for offsets slightly mixed with adding in a component to the tabline that was essentially just x amount of spaces as to bridge the gap between the last buffer and the window start of the next window. It shouldn't be massively complex since a lot of the behaviour is partially implemented but I think the utility is the biggest blocker.

akinsho:
I guess if the only thing the bufferline was showing was the buffers open in the visible windows then this mode might make sense, but it still leaves the question of what to do with non-top level splits e.g. a lower horizontal split or

mrjones2014:
Definitely understand the argument against something like this. The only real argument I have _for_ it is now that we can have a global status line (in a recent PR merged to Neovim), I'm looking for a different way to tell which file is open in which window (which, let me know if you have any alternative suggestions to solve this problem). I really like the global status line but this is one issue it causes.

>20 buffers open for example and 2 splits, what should happen in that case how do all the other buffers arrange themselves?

I'd think they'd just collapse behind each window to the right of it, if that makes sense.

> horizontal

[...]

PR #426, opened 2022-05-18
WIP: winbar mode

mrjones2014:
Fixes: #425

### DIFF
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -314,7 +314,12 @@ end
 ---@private
 function _G.nvim_winbarline()
   local current_buf = vim.api.nvim_win_get_buf(vim.g.statusline_winid)
-  if vim.api.nvim_buf_get_option(current_buf, "filetype") == "help" then
+  if
+    vim.tbl_contains(
+      require("bufferline.config").options.window_mode_ignore_filetypes or {},
+      vim.api.nvim_buf_get_option(current_buf, "filetype")
+    )
+  then
     return ""
   end
 
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -120,7 +120,7 @@ end
 local function toggle_bufferline()
   local item_count = config:is_tabline() and utils.get_tab_count() or utils.get_buf_count()
   local status = (config.options.always_show_bufferline or item_count > 1) and 2 or 0
-  if vim.o.showtabline ~= status then
+  if (not config:is_winbarline()) and vim.o.showtabline ~= status then
     vim.o.showtabline = status
   end
 end
@@ -312,6 +312,11 @@ function _G.nvim_bufferline()
   return bufferline()
 end
 
+---@private
+function _G.nvim_winbarline()
+  return bufferline()
+end
+
 ---@param conf BufferlineConfig
 function M.setup(conf)
   if not utils.is_current_stable_release() then
@@ -328,7 +333,11 @@ function M.setup(conf)
   highlights.set_all(preferences)
   setup_commands()
   setup_autocommands(preferences)
-  vim.o.tabline = "%!v:lua.nvim_bufferline()"
+  if config:is_winbarline() then
+    vim.o.winbar = "%!v:lua.nvim_winbarline()"
+  else
+    vim.o.tabline = "%!v:lua.nvim_bufferline()"
+  end
   toggle_bufferline()
 end
 
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -96,7 +96,6 @@ end
 --- @return string
 local function bufferline()
   local conf = config.get()
-  local tabs = tabpages.get()
   local is_tabline = conf:is_tabline()
   local components = is_tabline and tabpages.get_components(state) or buffers.get_components(state)
 
@@ -104,7 +103,7 @@ local function bufferline()
   --- state is not actually set till after sorting and component creation is done
   state.set({ current_element_index = get_current_index(state) })
   components = not is_tabline and groups.render(components, sorter) or sorter(components)
-  local tabline, visible_components = ui.render(components, tabs)
+  local tabline, visible_components, segments = ui.tabline(components, tabpages.get())
 
   state.set({
     --- store the full unfiltered lists
@@ -113,7 +112,7 @@ local function bufferline()
     components = filter_invisible(components),
     visible_components = filter_invisible(visible_components),
   })
-  return tabline
+  return tabline, segments
 end
 
 --- If the item count has changed and the next tabline status is different then update it
@@ -323,7 +322,7 @@ function _G.nvim_winbarline()
   return bufferline()
 end
 
----@param conf BufferlineConfig
+---@param conf BufferlineConfig?
 function M.setup(conf)
   if not utils.is_current_stable_release() then
     utils.notify(
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -314,7 +314,6 @@ end
 ---@private
 function _G.nvim_winbarline()
   local current_buf = vim.api.nvim_win_get_buf(vim.g.statusline_winid)
-  print(vim.g.statusline_winid)
   if vim.api.nvim_buf_get_option(current_buf, "filetype") == "help" then
     return ""
   end
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -314,6 +314,12 @@ end
 
 ---@private
 function _G.nvim_winbarline()
+  local current_buf = vim.api.nvim_win_get_buf(vim.g.statusline_winid)
+  print(vim.g.statusline_winid)
+  if vim.api.nvim_buf_get_option(current_buf, "filetype") == "help" then
+    return ""
+  end
+
   return bufferline()
 end
 
--- a/lua/bufferline/config.lua
+++ b/lua/bufferline/config.lua
@@ -573,6 +573,7 @@ local function get_defaults()
     ---@type BufferlineOptions
     options = {
       mode = "buffers",
+      window_mode_ignore_filetypes = { "help", "NvimTree" },
       themable = true, -- whether or not bufferline highlights can be overriden externally
       numbers = "none",
       buffer_close_icon = "",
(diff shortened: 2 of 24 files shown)


mrjones2014:
@akinsho super early stages and super simple, but let me know your initial thoughts. 

The thing I can't figure out is how to ensure that the highlighted/active buffer in the line is for the current window. It seems to mostly work but I think it just depends on when it gets redrawn. You can get into a state where it's showing an active buffer that's the active buffer for a different window if you keep moving around rapidly.

lua/bufferline.lua
```diff
@@ -312,6 +312,11 @@ function _G.nvim_bufferline()
   return bufferline()
 end
 
+---@private
+function _G.nvim_winbarline()
+  return bufferline()
```
akinsho, line 317: @mrjones2014 I think you can't rely on just rendering the bufferline as is, since this will just draw e.g. 30 buffers in each window bar which is messy and not very clear, I think what you'll want is only subset of what happens inside of the `bufferline` call to just call `ui.tabline` with only the one relevant buffer inside the current window.

akinsho: You'll need a new function like winline or something that call most of the same logic as bufferline but you need to ensure that it doesn't try and process all the buffers open in nvim since that will be wasteful as that logic will be happening across `x` number of window *every* time the tabline or statusline are redrawn

mrjones2014: So I actually wanted to show the full buffer list in each window, but just have the one shown in the "active" state to be the active buffer for the window. Maybe we can add some caching if necessary?

But for what it's worth, I tested it with tons of buffers and tons of windows open and it's plenty plenty fast enough.

mrjones2014:
How do I fix this issue where the current buffer is sometimes partially offscreen? Here the active buffer is the one cut off on the left.

image

lua/bufferline.lua
akinsho: @mrjones2014 I really don't see the point of that, why have all buffers visible above all the windows, that's seems so noisy from a UI perspective and doesn't really add any value? I really don't like that particular idea, apologies if any of this comes off as rude, definitely not intended that way.

Can you explain why that would be useful? The screenshot you've shown with the three windows each with a bufferline doesn't really look particularly nice, granted that's subjective, but I also think with the performance hit of N * Window count of the bufferline it needs a particularly good argument.

akinsho:
@mrjones2014 I've commented about this approach and I believe it's a side effect of trying to stuff the buffers above every window, which I really do not like tbh. I really think it should just be the active buffer above the window. It's likely because when determining available space atm, the tabline looks at the entire editor's size, not just the window size.

Regarding this all tabs UI, please consider other peoples usages of windows 
image
A lot of people I know use window in the above manner e.g. they have several open in complex layouts on big screens, this sort of all tabs above all layout will not make sense in these sorts of layouts.

lua/bufferline/config.lua
```diff
@@ -573,6 +573,7 @@ local function get_defaults()
     ---@type BufferlineOptions
     options = {
       mode = "buffers",
+      window_mode_ignore_filetypes = { "help", "NvimTree" },
```
akinsho: I generally avoid anything like this in this plugin because people strongly differ on what they do or do not want to see or which plugins they use. I wouldn't offer any items here and tbh think these sorts of changes should come as additional PRs to follow to keep the scope of this simpler

akinsho:
I tweaked your PR locally towards what I had been thinking, and it looks more like

image


```diff
diff --git a/lua/bufferline.lua b/lua/bufferline.lua
index e80f5ba..44d99d1 100644
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -19,6 +19,8 @@ local commands = lazy.require("bufferline.commands")
 local tabpages = lazy.require("bufferline.tabpages")
 --- @module "bufferline.constants"
 local constants = lazy.require("bufferline.constants")
+--- @module "bufferline.window"
+local window = lazy.require("bufferline.window")
 --- @module "bufferline.highlights"
 local highlights = lazy.require("bufferline.highlights")
 
@@ -97,12 +99,16 @@ end
 local function bufferline()
   local conf = config.get()
   local is_tabline = conf:is_tabline()
-  local components = is_tabline and tabpages.get_components(state) or buffers.get_components(state)
+  local is_winline = conf:is_winbarline()
+  local components = is_tabline and tabpages.get_components(state)
+    or is_winline and window.get_components(state)
+    or buffers.get_components(state)
 
   --- NOTE: this cannot be added to state as a metamethod since
   --- state is not actually set till after sorting and component creation is done
   state.set({ current_element_index = get_current_index(state) })
-  components = not is_tabline and groups.render(components, sorter) or sorter(components)
+  components = (not is_tabline and not is_winline) and groups.render(components, sorter)
+    or sorter(components)
   local tabline, visible_components, segments = ui.tabline(components, tabpages.get())
 
   state.set({
@@ -316,8 +322,8 @@ function _G.nvim_winbarline()
   local current_buf = vim.api.nvim_win_get_buf(vim.g.statusline_winid)
   if
     vim.tbl_contains(
-      require("bufferline.config").options.window_mode_ignore_filetypes or {},
-      vim.api.nvim_buf_get_option(current_buf, "filetype")
+      config.options.window_mode_ignore_filetypes or {},
+      api.nvim_buf_get_option(current_buf, "filetype")
     )
   then
     return ""
```
(NOTE: the diff is missing the window module which I added, but it just returns one buffer component)

I think this is a cleaner look and clearer, although the best version of this would just be the tabline up top but now with four less buffers since they are already shown in the winbar.

Although, I think with the behaviour of `WinSeparator` the background of the bar is not flush with the winseparator, so it looks less than ideal IMO 🤷🏾‍♂️ but could be still nice for others.

mrjones2014:
We can agree to disagree haha  :) 

I think I'll just implement my own in my dotfiles then. Feel free to use this PR if you'd like to commit your modifications here.

akinsho:
@mrjones2014 just to clarify, the key point that you wanted here is to see the bufferline duplicated above each window? Is there something I'm missing about why? I genuinely am curious if there's some use case I haven't considered? Is it just more aesthetically appealing to you that way?

mrjones2014:
For me, I often cycle through buffers in a window, while having multiple split windows. I'd want to see all the buffers that I can switch between in the tab line above each window, it just looks better and makes more sense to me.

dominicpalmer:
@mrjones2014 - I'm interested in this use case and I think many others would be too. I'd like to make the case for reopening the PR. 

@akinsho - The functionality being sought here is something that mirrors how VSCode's editors work. VSCode editors are basically windows, except each one has it's own tabline up top. The files in a window's tabline can only be viewed within that window. Example: two editors in vertial split, where one has three files 'a', 'b', 'c', loaded and available to view in that window, and the other has three different files 'x', 'y', 'z'. You can't view file 'a' in editor 2, and you can't view file 'x' in editor one. To do so, you would need to focus on a given editor, and open the file you want _in that editor_.

image

I find this to be the most intuitive way to work with files and windows in a text editor. It's also the most common - Vim is the exception to the rule in how it handles these concepts. Many like Vim's way, but some don't. Bufferline itself is a plugin that adds abstractions over the notion of the buffer and the tab to make things more intuitive.

I think it's only a matter of time before others want this functionality to be replicated now that winbar is merged and usable. Can we not put this behind an experimental setting and allow users to make the decision on whether their use case would suit it?

akinsho:
@dominicpalmer please read https://github.com/akinsho/bufferline.nvim/issues/348#issuecomment-1082213988 comment I made on the initial issue for this functionality.

What you are describing is vscodes concept of an `EditorGroup`, each editor group is almost a unique instance of vscode in that the buffers it contains (as you say) are unique to the editor group. *This concept does not exist in neovim* without extending this plugin to add this functionality entirely to the editor, which is quite frankly a completely separate plugin.

If you are willing to build this plugin, I would happily expose an API to integrate with it.

What this PR proposed is not the same of as what you are suggesting. This PR proposes *duplicating* the exact same tabline in all windows so to use your above example you would not see `a`, `b`, `c`, you would see `a`, `b`, `c`, `x`, `y`, `z` in *all windows*

dominicpalmer:
Understood. I know its not possible to directly implement the ways in which VSCode and emacs expose files and windows, but there may be avenues to explore in making Vim appear to do so.

Would it be possible to keep track of the nvim windows in which files were first loaded? And with that information, render contextual bufferlines for each split (vertical or horiztonal) that includes/excludes files as appropriate, depending on whether they have been loaded there by the user?

Example: 

`vim test.lua`. One file is loaded into a buffer and is seen in the single Neovim window. Bufferline for this window could show that file as a tab. `vsplit`. The same file is visible in both splits. A second bufferline for the split also shows the file. 

Then with focus in the split, open a different file, `test2.lua`. Because it was opened in the split, only the bufferline for that split makes the file visible as a tab, and the first splits bufferline is unchanged.

akinsho:
@dominicpalmer I've a very long time essentially on the concept of buffers & tabs in nvim, and how they can be represented in a way that potentially is closer to most other editors. I highly recommend if you are truly interested in this experimenting with this idea, I think you will find that it is much more involved to try to build this system on top of neovim, but I think possible. It's not a matter of a few functions, but really I think a whole system/plugin. There are so many edge cases such as how sessions are handled, how plugins that manipulate windows and buffers will interact with this "localized" windows etc.

Without diving too deeply into this, I believe the core logic of implementing some sort of separated editor/window local files is not the remit of a UI buffer/tabline plugin. This plugin is already way bigger than I wanted it to be, and I definitely don't want it growing into also being an editor state management plugin.

I should clarify I'm not against the idea of having something like editor groups in neovim, I think this would be great I don't know if there is appetite for it upstream, but my thinking is that someone needs to create a robust plugin that does this editor state management thing and once that exists it can be used here to power the bufferline view. I do not believe it is bufferline's responsibility and by proxy mine to do this.

mrjones2014:
@dominicpalmer I ended up implementing something similar as a fork of barbar.nvim if you're interested: https://github.com/mrjones2014/winbarbar.nvim

PR #439, opened 2022-06-02
fix(highlights): only set cterm values if not termguicolors

### LINKED ISSUE #433: [Bug]: failed setting highlight with color_icons = false option
### What happened?

Bufferline throws a `Failed setting BufferLineDevIconDefault  highlight, something isn't configured correctly: 'fg' is not a valid color` or some variation of this with a different highligh group. I may be wrong, but I suspect its the `ctermfg = not color_icons and "fg" or colors.get_color({` line in `ui.lua` as `"fg"` doesn't seem to be a valid value for `ctermfg` in `nvim_set_hl`.

No errors

### Additional Information

...

### commit

latest

akinsho:
Don't use `fg` for cterm colours since this will break if no colour was set for the cterm of a theme or more specifically for the Normal hl group

@rudotriton does this PR fix your issue, will probably break stuff for someone else 🤷🏿 

fixes #433

### DIFF
--- a/lua/bufferline/highlights.lua
+++ b/lua/bufferline/highlights.lua
@@ -64,9 +64,6 @@ local keys = {
   guibg = "background",
   guifg = "foreground",
   default = "default",
-  ctermfg = "ctermfg",
-  ctermbg = "ctermbg",
-  cterm = "cterm",
   foreground = "foreground",
   background = "background",
   italic = "italic",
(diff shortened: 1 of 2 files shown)


rudotriton:
it does fix it

akinsho:
@rudotriton can you try again I've changed the solution a bit so need to check it still works

rudotriton:
@akinsho the solution still work, thanks

PR #442, opened 2022-06-04
fix(tabs): explicitly convert segments text attribute to string

horseinthesky:
Explicitly convert the text attribute of each segment to string when calculating the component size. Some values can be numbers.
```
{
  highlight = "BufferLineBackground",
  text = "config.go"
}
{
  text = " "
}
{
  highlight = "BufferLineSeparator",
  text = "▕"
}
{
  highlight = "BufferLineBackground",
  text = " "
}
{
  highlight = "BufferLineBackground",
  text = " "
}
{
  attr = {
    __id = "number"
  },
  highlight = "BufferLineNumbers",
  text = 1
}
```

### DIFF
--- a/lua/bufferline/ui.lua
+++ b/lua/bufferline/ui.lua
@@ -122,7 +122,7 @@ local function get_component_size(segments)
   local sum = 0
   for _, s in pairs(segments) do
     if has_text(s) then
-      sum = sum + strwidth(s.text)
+      sum = sum + strwidth(tostring(s.text))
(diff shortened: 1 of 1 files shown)


akinsho:
Thanks @horseinthesky I've now merged it into `dev` I changed the base since I like all changes to go into `dev` first to settle and make sure nothing is broken then eventually merge into `main`, not that that always works, probably need more testers on `dev`

PR #447, opened 2022-06-09
refactor(ui): allow trunctation to follow direction

### LINKED ISSUE #413: Don't center tabs when scrolling
I always lose track of the buffer I am at because buffers get shifted all the time.

It is easier to show than explain.

Here's what it looks like in `bufferline.nvim`:
Peek 2022-05-11 08-33

Here's how it looks like in `barbar`:
Peek 2022-05-11 08-36

An option to change this setting would be nice.

akinsho:
This PR allows the current buffer to navigate till the end of the visible buffer line rather than always staying centered.

This PR fixes #413

### DIFF
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -221,7 +221,23 @@ local function setup_autocommands(conf)
 
   api.nvim_create_autocmd("BufEnter", {
     pattern = "*",
-    callback = function()
+    callback = function(args)
+      for index, component in ipairs(state.components) do
+        if component.id == args.buf then
+          state.set({ current_element_pos = index })
+        end
+      end
+      handle_group_enter()
+    end,
+  })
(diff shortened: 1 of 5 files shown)


akinsho:
This is hard and probably needs a re-imagined mechanism for showing buffers

tmerse:
Would love to see this happen. Any idea if this will land at some point?

akinsho:
@tmerse I have no idea how to finish this PR, and in its current state it's probably no longer relevant. If you have any suggestions on implementation or can contribute a PR please feel free.

tmerse:
Thanks for the quick reply. I am pretty swamped with work for now and the forseeable future.
I switched back from barbar to bufferline and I like else in bufferline better except it's hard for me to visually track the buffer, but it's a minor pain point.
If I manage to find the time, I'll have a look and see if I can come up with sth.

Issue #438, opened 2022-06-02
[Feature Request]: Commands `BufferlineClose*` close modified buffers


Shatur:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

Commands `BuferlineCloseLeft`, `BuferlineCloseRight` and `BufferlinePickClose` close modified buffers.

### What did you expect to happen?

I would expect to have an error message like "Unable to close modified buffer, add ! to override". So it would be nice to have the current behavior with bang and the expected behavior without it.

### Config

```lua
require("bufferline").setup({})
```

akinsho:
@Shatur under the hood the API function used is called with force set to true which is my personal preference I don't mind there being a bang as an option to set this value but personally like the default behaviour and am not inclined to change it. If you would like to raise a PR to tweak this I'd be happy to review but I'm definitely not keen on adding this functionality myself since I already personally like the current behaviour

akinsho:
This is now fixed on `dev` by #563

Issue #279, opened 2021-12-24
How do you configure bufferline to not close unsaved files?


ninjarogue:

ninjarogue:
Is there a way to do this with bufferline?

ninjarogue:
Okay so I tried `close_command = 'bdelete %d'` but it's still closing my buffers even if there are unsaved changes

akinsho:
@ninjarogue I believe this is really a question about the command in nvim and less specific to this plugin, so I think searching nvim help and the web would be an easier place to find an answer. On my side I autosave on leaving a file so have never had this problem. You can try setting `confirm` see `:h confirm`. Alternatively, by default nvim won't close a file using `bdelete <num>` if no `!` is passed to the `bdelete!` command, so it should already be working and might have something to do with how your nvim is configured.

PR #451, opened 2022-06-13
Prevent unsaved buffers to be closed without confirmation

jpoppe:
Removed default force flag (!) from `bdelete` and 'api.nvim_buf_delete' commands.

I don't know if it is intentional but IMHO force deleting buffers  is not a user friendly default.

I lost quite some work today by using the default configuration
and using the `BufferLineCloseLeft` and `BufferLineCloseRight`
and not expecting it would close unsaved buffers without confirmation.

---

If the intended behavior is `force` deleting buffers please feel free to close this pull request and maybe you could add a warning to the README that it force deletes buffers, thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,10 +212,10 @@ require('bufferline').setup {
   options = {
     mode = "buffers", -- set to "tabs" to only show tabpages instead
     numbers = "none" | "ordinal" | "buffer_id" | "both" | function({ ordinal, id, lower, raise }): string,
-    close_command = "bdelete! %d",       -- can be a string | function, see "Mouse actions"
-    right_mouse_command = "bdelete! %d", -- can be a string | function, see "Mouse actions"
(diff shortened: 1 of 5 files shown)


jpoppe:
See also: #279

akinsho:
@jpoppe this relates also to #438 where there is a similar request. So this was definitely by choice since current behaviour is definitely my preference. As I mention though I'm happy to have this be configurable and my suggestion in the issue I linked was that if a `bang` is passed in the command it should force a deletion otherwise it shouldn't.

With the addition of lua commands this is now much easier since you only need to check the args table passed in to the `CloseRight/Left` commands and see if `bang = true` if so pass that to the delete function

jpoppe:
Thanks for your fast reply! This was indeed a quick fix on my side and it would be of course nicer if it's configurable.more flexible, was just trying this plugin and did not expect this behavior. And I missed #438 in my searches 🤔 .

akinsho:
This feature has now been merged into the dev branch

Issue #456, opened 2022-06-19
[Feature Request]: Remove vimscript wrapper for click support


nullchilly:
### What?

```
function Click(num, clicks, button, flags)
	print(num, clicks, button, flags)
end

vim.o.tabline = "%1@v:lua.Click@ Click"
```
This works

### Why?

No more https://github.com/akinsho/bufferline.nvim/blob/main/autoload/nvim_bufferline.vim

akinsho:
@nullchilly have you tried the example in the snippet that you shared? Because I did a while ago, and it doesn't work. So neovim upstream have fixed it. It is still required to use a vim script wrapper

nullchilly:
Yes it didn't work for 0.5

akinsho:
Can you raise a PR? and I'll test it, if it works then I'll happily merge it 👍🏾

nullchilly:
It will be merged to main when neovim 0.8 released

akinsho:
Ah, thanks for the heads-up @nullchilly didn't realize it was 0.8 only, well it's in the dev branch for now

nullchilly:
Fortunately I didn't break dozen people's workflow D:

akinsho:
Gonna close this out since the work is technically done though 👍🏿

PR #457, opened 2022-06-19
refactor(handle_click): remove vim script wrapper

nullchilly:
For some reason I failed to make them work with `v:lua.require'bufferline'.` Other than that, works for me :) 
Oops all checks have failed lol. Do I need to rewrite the tests too

Fixes: https://github.com/akinsho/bufferline.nvim/issues/456


lua/bufferline/ui.lua
```diff
@@ -145,7 +145,7 @@ function M.make_clickable(func_name, id, component)
   -- v:lua does not support function references in vimscript so
   -- the only way to implement this is using autoload vimscript functions
   component.attr = component.attr or {}
-  component.attr.prefix = "%" .. id .. "@nvim_bufferline#" .. func_name .. "@"
+  component.attr.prefix = "%" .. id .. "@v:lua.nvim_bufferline_" .. func_name .. "@"
```
akinsho, line 148: There are probably more places than just here probably where this needs to also be changed, please have a look to make sure you got them all

lua/bufferline/commands.lua
```diff
@@ -109,7 +109,7 @@ local cmds = {
 ---Handler for each type of mouse click
 ---@param id number
 ---@param button string
-function M.handle_click(id, button)
+function _G.nvim_bufferline_handle_click(id, _, button)
```
akinsho, line 112: This name is probably quite likely to cause a namespace clash. I suggest something much more secure if we are going to pollute the global namespace e.g. create a global table and then put a handle click function inside it e.g.

```lua
_G.___bufferline_private = _G.__bufferline_private or {} -- to guard against reloads
_G.___bufferline_private.handle_click = ...

```

akinsho (changes requested):
@nullchilly thanks for the PR left some comments about improving the naming to make the global variables less easy to accidentally override or discover. 

Also, you will need to update the tests probably you just need to change the names in the tests also

lua/bufferline/ui.lua
nullchilly: I searched them through telescope through the keyword `nvim_bufferline#` as that's the only way to call the vimscript I think?

lua/bufferline/commands.lua
nullchilly: Where do I put these global variables, at the top?

akinsho: @nullchilly you can put it near the top bufferline.lua file after the imports and maybe add a comment about why it's there

lua/bufferline.lua
```diff
@@ -22,6 +22,10 @@ local constants = lazy.require("bufferline.constants")
 --- @module "bufferline.highlights"
 local highlights = lazy.require("bufferline.highlights")
 
+-- @v:lua@ in tabline only support global function so this is
+-- the only way without autoload vimscript functions
```
akinsho, line 26:
```suggestion
-- @v:lua@ in the tabline only supports global functions, so this is
-- the only way to add click handlers without autoloaded vimscript functions
```

lua/bufferline.lua
```diff
@@ -22,6 +22,10 @@ local constants = lazy.require("bufferline.constants")
...
+_G.___bufferline_private = _G.__bufferline_private or {} -- to guard against reloads
```
akinsho, line 27:
```suggestion
_G.___bufferline_private = _G.___bufferline_private or {} -- to guard against reloads
```

akinsho (approved):
LGTM 👍🏿 thanks 🚀

nullchilly:
Since this is unmerged is it possible to reopen the pupl request?

akinsho:
@nullchilly I added the merge commit to the version 3 branch which i will merge when 0,8 is stable

PR #458, opened 2022-06-20
v3.0.0

akinsho:
DNM till nvim 0.8 is stable


nullchilly:
https://github.com/nvim-lualine/lualine.nvim/commit/0ac721d37637ab1903b66822fd1a1ee9d08013f0#diff-8343eceafa3f37676f18bc9653f752cffd9d1699427c28369a1fa0b747f2dd88R253

It seems like we don't need to use global lua function anymore

PR #459, opened 2022-06-20
feat: add option/ability to omit diagnostics entirely

vanarok:
2022-06-20T18:05:36,535511116+04:00
With the new diagnostics_pure=true option two things happen: 
1. The diagnostics count is hidden 
2. The spacing  after diagnostics is hidden.

### DIFF
--- a/lua/bufferline/diagnostics.lua
+++ b/lua/bufferline/diagnostics.lua
@@ -178,6 +178,11 @@ function M.component(context)
   local diag_highlight = highlights[diagnostics.level .. "_diagnostic"]
     or highlights.diagnostic
     or ""
+
+  if config.options.diagnostics_pure then
+    indicator = ""
+  end
+
   return {
     text = indicator,
     highlight = diag_highlight,
--- a/lua/bufferline/ui.lua
+++ b/lua/bufferline/ui.lua
@@ -537,7 +537,7 @@ function M.element(state, element)
     set_id(name, components.id.name),
     spacing({ when = name, highlight = curr_hl.buffer.hl }),
     set_id(diagnostic, components.id.diagnostics),
-    spacing({ when = diagnostic }),
+    spacing({ when = diagnostic and not config.options.diagnostics_pure }),
     right_space,
     suffix,
     spacing({ when = suffix }),


lua/bufferline/ui.lua
```diff
@@ -460,8 +460,8 @@ end
 ---@param opts SpacingOpts?
 ---@return Segment
 local function spacing(opts)
-  opts = opts or { when = true }
-  if not opts.when then
+  opts = opts or { when = true } and { pure = true }
```
akinsho, line 463: This is an implementation detail of diagnostics that should not appear here at all, I would remove all of this

lua/bufferline/diagnostics.lua
```diff
@@ -178,6 +178,11 @@ function M.component(context)
   local diag_highlight = highlights[diagnostics.level .. "_diagnostic"]
     or highlights.diagnostic
     or ""
+
+  if config.options.diagnostics_pure then
```
akinsho, line 182: I don't believe a new option is necessary @vanarok rather than doing this, I recommend changing this

https://github.com/akinsho/bufferline.nvim/blob/main/lua/bufferline/diagnostics.lua#L166-L170

To

```lua
  local indicator = ""
  if user_indicator and type(user_indicator) == "function" then
    local ctx = { buffer = element, tab = element }
    indicator = user_indicator(diagnostics.count, diagnostics.level, diagnostics.errors, ctx)
  elseif user_indicator == nil then -- If nothing is specified at all use the default
    indicator = " (" .. diagnostics.count .. ")"
  end
```

lua/bufferline/ui.lua
```diff
@@ -537,7 +537,7 @@ function M.element(state, element)
     set_id(name, components.id.name),
     spacing({ when = name, highlight = curr_hl.buffer.hl }),
     set_id(diagnostic, components.id.diagnostics),
-    spacing({ when = diagnostic }),
+    spacing({ when = diagnostic, pure = config.options.diagnostics_pure }),
```
akinsho, line 540: This can just be changed to 
```suggestion
    spacing({ when = diagnostic and diagnostic ~= "" }),
```

akinsho (changes requested):
@vanarok thanks for taking this on, I've recommended some changes because I think this can all be achieved without adding yet another config option. A user should simply specify `false` and there will be no diagnostics shown unless they pass in a function or nothing i.e. `nil`

akinsho:
@vanarok please explain what filtering you are referring to? I don't really understand what you are filtering or how that relates to this PR

For context, I definitely do not want any new config option if possible there are already like 80 which is way too many. I also definitely do not want the implementation details of such a thing getting into the ui rendering layer in that way, so we need a solution that doesn't involve that (that is a blocker)

akinsho:
Ah, it seems maybe you mean that you are using the indicator to control what diagnostics appear in the bufferline, this is not how that was intended to be used to be honest. One solution is to allow the user indicator to return multiple values and use one to indicate emptiness, but tbh I don't really like the implicit usage of a function for creating and indicator as a filter function

vanarok:
> Ah, it seems maybe you mean that you are using the indicator to control what diagnostics appear in the bufferline, this is not how that was intended to be used to be honest. One solution is to allow the user indicator to return multiple values and use one to indicate emptiness, but tbh I don't really like the implicit usage of a function for creating and indicator as a filter function

`
                        diagnostics_pure = true,
                        diagnostics_indicator = function(count, level)
                            if level == 'error' then
                                return 'any text will be hidden with diagnostics_pure=true'
                            end
                        end,
`
That's what i mean.

akinsho:
@vanarok in the example above, it is unclear to me why a person would want to have no indicator but be returning text inside the diagnostic indicator but want it hidden?

akinsho:
Also, this is tangential since an option is a last resort to avoid anyway but the term `pure` is very vague, subject and not at all in keeping with the rest of the option naming. I would call it something much clearer and simpler like, `hide_diagnostic_indicator`

akinsho:
@vanarok I don't mind changing those if it allows us to avoid another option, but why return `true` from the indicator at all? Why use it at all instead of just setting `diagnostic_indicator = false`

vanarok:
And besides, the highlighting of the file name is also part of the diagnostic indicator. My original suggestion was to disable the count and diagnostic icons, the highlighting should not be disabled.

akinsho:
Ok I think we are going about this incorrectly because there's a sense of coupling the diagnostic indicator to stuff that isn't related to it. It is just a function to return a string of diagnostics, it isn't a mechanism for filtering. It also shouldn't affect the diagnostic highlights.

If it affects the diagnostic highlights e.g. removing the function removes the highlight, then that is a bug/problem to refactor. A user should be able to set false to the indicator and still get highlights for diagnostics for the buffer, since they should be unrelated.

The config example above I think shouldn't be a given because it represents a confusing set of options, e.g. a user is setting the value of a function called an indicator but setting another value to disable it. This is confusing and to be avoided, IMO.

I think a good starting place is to turn off the indicator, i.e. `false` and then see what the bugs or problems are with that approach. If the highlights aren't showing then that should be fixed since they shouldn't be connected.

If there are other issues feel free to raise them here if you don't know how to solve them but the ideal PR for this to my mind is the ability to explicitly disable the indicator and have your desired minimal aesthetic.

akinsho:
I've implemented what I mentioned here in #460, so I think this can be closed. You can now just set the indicator to false to achieve this, so there will be highlighting but no indicator

PR #465, opened 2022-07-03
fix(tabpages): use tabpagenr for closing tabpages

GabMus:
Hey, you might remember me working on tabpage filtering a while ago.

Since then I've discovered a bug where basically bufferline tried to close tabpages using the tabpage handle instead of the tabpagenr.

I finally got the time to find the code responsible for this and fix it! As you can see it was fortunately a very simple fix.


akinsho (approved):
LGTM @GabMus thanks for contributing (again ❤️ )

Issue #466, opened 2022-07-05
[Feature Request]: Jumping to the ordinal buffer number


gerazov:
### What?

Great plugin :sunglasses: 

Add a way to jump to the ordinal buffer number in bufferline.nvim.

I'm using:

```vim
nnoremap <silent><a-1> <Cmd>BufferLineGoToBuffer 1<CR>
nnoremap <silent><a-2> <Cmd>BufferLineGoToBuffer 2<CR>
nnoremap <silent><a-3> <Cmd>BufferLineGoToBuffer 3<CR>
nnoremap <silent><a-4> <Cmd>BufferLineGoToBuffer 4<CR>
nnoremap <silent><a-5> <Cmd>BufferLineGoToBuffer 5<CR>
nnoremap <silent><a-6> <Cmd>BufferLineGoToBuffer 6<CR>
nnoremap <silent><a-7> <Cmd>BufferLineGoToBuffer 7<CR>
nnoremap <silent><a-8> <Cmd>BufferLineGoToBuffer 8<CR>
nnoremap <silent><a-9> <Cmd>BufferLineGoToBuffer 9<CR>
```

But this only jumps based on the visible buffers in the bufferline. I'd like to jump to the first file opened when I do `<a-1>`

### Why?

I'm coming from [vim-airline](https://github.com/vim-airline/vim-airline) and there one could jump to the ordinal number of the buffer, i.e. the absolute number. 

I use this often for jumping to the first `org` file opened, which is always the same for me.

gerazov:
Ok - had to dig through but found it #210 thanks :pray: 

And here's the code here for reference:
```vim
nnoremap <silent><leader>1 <cmd>lua require("bufferline").go_to_buffer(1, true)<cr>
nnoremap <silent><leader>2 <cmd>lua require("bufferline").go_to_buffer(2, true)<cr>
nnoremap <silent><leader>3 <cmd>lua require("bufferline").go_to_buffer(3, true)<cr>
nnoremap <silent><leader>4 <cmd>lua require("bufferline").go_to_buffer(4, true)<cr>
nnoremap <silent><leader>5 <cmd>lua require("bufferline").go_to_buffer(5, true)<cr>
nnoremap <silent><leader>6 <cmd>lua require("bufferline").go_to_buffer(6, true)<cr>
nnoremap <silent><leader>7 <cmd>lua require("bufferline").go_to_buffer(7, true)<cr>
[...]

Issue #210, opened 2021-09-01
Recent changes to BufferLineGoToBuffer


lcrockett:
I was used to using `BufferLineGoToBuffer` based on the ordinal number of the buffer, regardless of it's visual position. However in commit `d343bc322e00e9b21bcb9b3c8bf6a989b64a92a2` this behaviour changed to jumping to a buffer based on it's visual position. It does make sense from one perspective, however it seems cumbersome to keep counting the how "many'th" buffer you'd like to jump to and only then picking it using `BufferLineGoToBuffer` (for times when `BufferLinePick` seems a tad bit slower then \<leader\>+\<number\>).

Is this change in behaviour permanent or does that functionality still remain in one way or another ?

Cheers on keeping up with the continuous stream of improvements to an excellent bufferline plugin.

akinsho:
@lcrockett the change was intentional and didn't offer an alternative because I'm trying to manage the proliferation of options in this plugin. Ultimately, I don't want there to be a bunch of options around this. The main issue is that counting visible buffers is an overhead, ideally there'd be an option to show these but atm that'd be complicated to do. In the short/medium term, there is a way round this.

You can create your own command using the `require'bufferline'.go_to_buffer(number, absolute)` absolute is a boolean that determines whether to use the absolute buffer position or the visible/relative one. You can then define a command similar to the existing one that does this.

lcrockett:
Perfect, works like a charmadillo. Cheers !

stumash:
@lcrockett mind posting the code you wrote? I want this too

lcrockett:
@stumash I don't have much code to show you, i've incorporated @akinsho's recommendation in my `which-key` configuration. See below:

```lua
    ['1'] = { '<cmd>lua require("bufferline").go_to_buffer(1, true)<cr>', 'which_key_ignore' },
    ['2'] = { '<cmd>lua require("bufferline").go_to_buffer(2, true)<cr>', 'which_key_ignore' },
    ['3'] = { '<cmd>lua require("bufferline").go_to_buffer(3, true)<cr>', 'which_key_ignore' },
    ['4'] = { '<cmd>lua require("bufferline").go_to_buffer(4, true)<cr>', 'which_key_ignore' },
    ['5'] = { '<cmd>lua require("bufferline").go_to_buffer(5, true)<cr>', 'which_key_ignore' },
[...]

PR #467, opened 2022-07-05
added absolute number jumping information

gerazov:
Added info about mapping jumping to absolute position using the `lua` API as per #210 and #466.


akinsho (approved):
LGTM 👍🏿 thanks

gerazov:
np - the least I could do :pray:

PR #483, opened 2022-07-21
docs(bufferline.txt): add missing hint highlights

tobealive:
Had an occasional highlight that didn't played along and couldn't figure out whats wrong as it was missing in the help docs. 

Let us make the `hint`-group as easy accessible as the other ones 😊


akinsho (approved):
Awesome, thanks for adding this 👍🏿

Issue #430, opened 2022-05-22
[Enhancement]: Add deduplication in tab mode


wassimk:
### What happened?

Unique buffer names are not working for me when using `mode` `tabs`.  

If I set the `mode` to `buffers`, it works great.

### What did you expect to happen?

Same configuration with `mode` = `buffers`. This works as expected.

![image](https://user-images.githubusercontent.com/2261839/169719506-bb6add0f-5974-4f05-80cf-a49d98bd7f56.png)

Unexpected result with `mode` = `tabs`. These are the same files but don't get deduplicated.

![image](https://user-images.githubusercontent.com/2261839/169719518-5e0f7e9a-dc58-4a58-a62c-8f78d601aceb.png)

### Additional Information

``` lua
require('bufferline').setup {
  options = {
    mode = 'tabs',
    show_buffer_close_icons = true,
    show_close_icon = false,
    always_show_bufferline = false,
    separator_style = 'slant',
    enforce_regular_tabs = false,
    offsets = {
      {
        filetype = 'NvimTree',
        text = nil,
      }
    }
  }
}
```

I saw issue #71 describing the same issue I have, but I have `enforce_regular_tabs` set to `false`.

akinsho:
Hi @wassimk this is actually not a bug, excerpt from the README

> 
> This will change the bufferline to a tabline it has a lot of the same features/styling but not all. A few things to note are
> 
>     * Diagnostics only show if the buffer with issues is the current window selected in that tab page
>     * Sorting doesn't work yet as that needs to be thought through.
>     * Grouping doesn't work yet as that also needs to be thought through.

So I primarily develop features with buffers in mind which is why I created this plugin but after a few recurrent issues asking for tabs I finally implemented this but explained at the time that it wasn't a 1:1 implementation and that some features need to be thought through for tabs and implemented for tabs, so I only implemented the simplest ones and I depend on tab users to add the remaining functionality having thought through how they would work best for tabs. e.g. #398 where a tab user recently implemented the custom filter for tabs.

I do not personally intend to do this work because I don't use tabs like this and depend on contributions for these. This shouldn't be too complex i.e. a matter of re-using the existing deduplication logic on the active tab but might need some tweaking

wassimk:
Thank you for your time in this response, @akinsho. I read similar in this repo and tried hard to read through the docs and old issues to assure myself I was genuinely asking about unexpected behavior. I read this same part of the readme and didn't process that it would include deduplication. 

[...]

PR #485, opened 2022-07-24
feat(tabpages): support unique tab names

wassimk:
This implements unique names for tabpages.

I looked around for documentation to update, but I don't see anything to change. The feature will just work now for tabs like it does and is described for buffers.

**Here is a screenshot of some unique tabs:**

image

- _hammerspoon/init.lua_
- _config/nvim/init.lua_
- _config/nvim/lua/wassim/cmp/init.lua_

**For clarity, here is my current config:**

``` lua
require('bufferline').setup({
  options = {
    mode = 'tabs',
    show_close_icon = false,
    always_show_bufferline = false,
    separator_style = 'slant',
    enforce_regular_tabs = false,
    offsets = {
      {
        filetype = 'NvimTree',
        text = nil,
      },
    },
  },
})
```

Closes https://github.com/akinsho/bufferline.nvim/issues/430.

### DIFF
--- a/lua/bufferline/tabpages.lua
+++ b/lua/bufferline/tabpages.lua
@@ -8,7 +8,7 @@ local config = lazy.require("bufferline.config")
 ---@module "bufferline.constants"
 local constants = lazy.require("bufferline.constants")
 --- @module "bufferline.duplicates"
-local duplicates = require("bufferline.duplicates")
+local duplicates = lazy.require("bufferline.duplicates")
 ---@module "bufferline.diagnostics"
 local diagnostics = lazy.require("bufferline.diagnostics")
 ---@module "bufferline.utils"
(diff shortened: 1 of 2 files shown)


lua/bufferline/tabpages.lua
```diff
@@ -7,6 +7,8 @@ local pick = lazy.require("bufferline.pick")
 local config = lazy.require("bufferline.config")
 ---@module "bufferline.constants"
 local constants = lazy.require("bufferline.constants")
+--- @module "bufferline.duplicates"
+local duplicates = require("bufferline.duplicates")
```
akinsho, line 11:
```suggestion
local duplicates = lazy.require("bufferline.duplicates")
```

`lazy.require` lazy loads lua modules to defer the cost of requiring so many modules till the first time that they are accessed, e.g. `module.field` rather than when this file is first read. That's the reason why I annotate them as `---@module` this allows sumneko to know what the underlying lazy loaded module is

akinsho (approved):
LGTM 🚀 just made one comment to tweak otherwise this is G2G

PR #495, opened 2022-08-02
refactor(ui): update offset text overlength handling

tobealive:
This should fix some jumpiness when the offsets text length exceeds the width of the window that the offset is applied for. (Currently, the part of the bufferline that's containing the buffers moves inside the offset if it has overlength). Additionally, the PR adds an indicator that there is more text.

I guess there could be a reason for doing it the way that it is done right now. And I'll probably find out soon. 

But giving you something visual what I'm changing with this PR is probably the easiest, so I'll add two quick and dirty 10sec videos below. For this example, I'm using the path function from the readme that is using a simple `vim.fn.getcwd()` to get the path in the offset.

<details>
<summary>Current:</summary>
https://user-images.githubusercontent.com/34311583/182484665-42084722-dddb-4429-b9f6-cd1eaad8c66f.mp4
</details>

<details>
<summary>PR:</summary>
https://user-images.githubusercontent.com/34311583/182484787-39b06daa-9c1d-4643-9ad9-9bfcbb067da3.mp4
</details>

---

_Little additional obervations with this PR (nothing essential):_
In my usual config, I'm using a custom function to truncate the path in the offset. With this there's one additional thing I'm observing.
The current way of handling overlength sometimes results in some cryptic characters that' are added to the "`sub`bed" text. Sometimes even in crashes when resizing an offset with overlength.
Funnily this seems not to happen anymore when using the changes in the PR. The reasons for this are also yet to be discovered 🧐 (really no idea why the changes in the PR can have an effect on that). Most definitely I could have done better on that truncating function so that the current way of handling this won't result in cryptic chars and stuff.

### DIFF
--- a/lua/bufferline/offset.lua
+++ b/lua/bufferline/offset.lua
@@ -30,8 +30,8 @@ local function get_section_text(size, highlight, offset)
   else
     local text_size = fn.strwidth(text)
     local left, right
-    if text_size + 2 >= size then
-      text = text:sub(1, size - 2)
+    if text_size >= size then
+      text = text:sub(1, size - 1) .. "…"
       left, right = 1, 1
     else
       local remainder = size - text_size


tobealive:
Oh damn, it looks like the reason for this jumpiness it the icon I added to the beggining path 🤦🏾‍♂

akinsho:
@tobealive so I've hit this a few times in the plugin and a user actually submitted a truncation helper function which is in plugin utils, truncate by char or something it's probably a matter of applying that in the offset logic.

akinsho:
@tobealive to clarify the usage of that function it is intended to be used *instead of `string.sub`*. It's not going to solve any problems if the logic is still using sub because that method is not safe for truncating things that are not latin alphanumeric characters. I'm not keen on a super complex solution to this. I haven't much time at the moment so haven't dug into what exactly the issue is but the way the function works is you tell it what the max size is and it will truncate the function cell by cell keeping multibyte characters together until it hits the acceptable length in cell sizes.

tobealive:
Ahh, I'm getting you now. Thanks for clarifying. 

I thought your first comment should tell me something like:
"This is something that is already solved. You just need to use this in your config."

Unfortunately this is also not solving the issue. Using it in the `offsets.lua`s `get_section_text` function instead of lua's `sub` will actually make it even more jumpy (Video example below).

https://user-images.githubusercontent.com/34311583/183291098-bdfb6094-04fa-4cac-b79e-d9741ec7f985.mp4

Additionally, if one uses something like a leading icon in the offset text it will also be removed as soon as truncation happens.  


I think this might be a problem that has a very limited amount of solutions. The one I'm mentioning here might be one for it.
> I was able fix it now in a dynamically working way by counting the position of the characters in bytes and subtracting the currently used text size. But this needs [uga-rosa/utf8.nvim](https://github.com/uga-rosa/utf8.nvim)'s utf8.codes()

tobealive:
Following two solutions for both parts of the problem.

will fix the issue with jumpiness if there are leading utf8 characters:
https://github.com/tobealive/bufferline.nvim/commit/ac5f8c71b3201f47302f37b9b9501d8aa1f2aa54

will fix partial slicing of utf8 characters, but  still has some jumpiness:
https://github.com/tobealive/bufferline.nvim/commit/29acddca66f9616d057ec7530385fbeebc630b32

Trying to get one proper working piece out of those I'm either running into the problem that I'm stupid to find the mistakes I'm doing, being too time constraint to dig myself deeper into this. Mostly both.

But maybe you or someone can make a use of it if you want to tackle this.

akinsho:
I've merged #503 which should fix this issue, let me know if it somehow doesn't cover your issue

akinsho:
This is only live in `dev` which is where all commits go before release

PR #503, opened 2022-08-14
fix(offset): truncate long offset text with multibyte characters

akinsho:
This PR fixes #495 

@tobealive can you please try this PR out. It uses the utility function I mentioned to truncate the offset text specified, whether it includes multibyte characters. The issue you were having with that function comes down to the fact that the text was being changed to try and remove the extension, but this was initially intended for use with buffer names with extensions, so it ended up changing the text if using a directory, but that should be fixed now

image

### DIFF
--- a/lua/bufferline/constants.lua
+++ b/lua/bufferline/constants.lua
@@ -29,4 +29,6 @@ M.visibility = {
 
 M.FOLDER_ICON = ""
 
+M.ELLIPSIS = "…"
+
 return M
--- a/lua/bufferline/offset.lua
+++ b/lua/bufferline/offset.lua
@@ -31,8 +31,7 @@ local function get_section_text(size, highlight, offset)
   if type(text) == "function" then text = text() end
   text = text or string.rep(" ", size)
 
-  local left, right
-  local text_size = api.nvim_strwidth(text)
+  local text_size, left, right = api.nvim_strwidth(text), 0, 0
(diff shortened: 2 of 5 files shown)


akinsho:
There are a lot of unrelated changes here that I added just as a minor spring clean, the core change that was needed was to use `utils.truncate_name` and then fix the behaviour of that function to not `fnamemodify` the path using `:r:h` which mangles directories

tobealive:
Thanks @akinsho 🚀
It truncs what it should now, great work and nice cleanups 😊.

A thing that I notice with the truncation, is that my nvim locks when the minimal width of a window that an offset is applied to is reached. 
It probably won't be a frequent occurrence that users manually resize to the lowest width, but a situation that I sometimes find myself in is, that nvims internal windows can get squished to the lowest width on terminal window resize. Provoking such a situation also causes a lock.

akinsho:
@tobealive this is probably because there is a while loop in there somewhere, and it is probably missing an exit condition to handle minuscule values. Tbh kind of tapped out bufferline stuff for the moment as I need to focus elsewhere, feel free to raise an issue or a PR for it 🤷🏿. Need to focus elsewhere though, so won't be looking at it soon, probably.

tobealive:
Thanks for the info. I'll make some time to check for it in the coming days and do my best on it.

PR #507, opened 2022-08-14
fix: freeze with miniscule truncation values

tobealive:
- related to: https://github.com/akinsho/bufferline.nvim/pull/503#issuecomment-1214373879


akinsho (approved):
LGTM 👍🏾

tobealive:
your tip made it a nobrainer, it's your merrit. 

But I chanced it if you prefer a variable to not do a tripplet call on the width func. Out the top of my head I don't know if this would make a difference in lua.

akinsho:
@tobealive I was actually considering that as a refactor, I'm not sure how expensive that function is but it's being called a lot but it's not a huge deal I think

PR #508, opened 2022-08-15
Document highlight groups for bufferline tab numbers

HerCerM:
I noticed the following highlight groups are not documented: `numbers`, `numbers_visible` and `numbers_selected`. I had to look in the code to realize they exist. I believe it can be useful for other users to have them in the docs.


akinsho (approved):
Thanks for the contribution, LGTM 👍🏾

PR #519, opened 2022-08-22
fix(utils): improve path separator detection on windows

aarrwnh:
`vim.loop.os_uname()` output:
```
{
  machine = "x86_64",
  release = "10.0.19044",
  sysname = "Windows_NT",
  version = "Windows 10 Enterprise"
}
```

'Side-effect' of this is that path separator might look like pipe with some fonts (because of italics) now.


akinsho (approved):
@aarrwnh thanks for the PR 👍🏿 much appreciated

PR #520, opened 2022-08-23
docs(help): add missing comma to indicator config

sanders41:
In updating my config for the new indicator icon settings I was getting an error after copy and pasting from the docs. I noticed it was because of a missing comma in the docs.


akinsho (approved):
LGTM 👍🏿

akinsho:
@sanders41 this LGTM, as a side note though I generally strongly recommend against copy pasting any code in the docs, I really don't check to ensure they are safe for copy pasting because that adds way too much extra burden on me to checking every block. I also *want users to read through the config carefully*

PR #523, opened 2022-08-24
bugfix(config): reset highlights on background change

### LINKED ISSUE #522: [Bug]: Highlighting do not reload on changing background

### What happened?

Highlighting stop reloading on changing background after f248437e31b97f2a91173ee770b181b8d8ccdd42
To reproduce just call `:set background=light` or `:set background=light` on a theme that supports light and dark mode. For example, https://github.com/Shatur/neovim-ayu

Highlighting changes on changing background setting.

### Config

No configuration required.

### Additional Information

_No response_

### commit

f248437e31b97f2a91173ee770b181b8d8ccdd42

akinsho:
When the colorscheme changes for any reason, we should essentially reset highlighting back to the user's initial config and reload all highlights

fixes #522 

@Shatur can you try this out? seems to fix it on my end

### TODO
- [ ] write much needed tests for colorscheme changes


Shatur:
Works for me, thank you!

avegancafe:
FWIW, this fixes the same problem I had as well. Not quite sure how to force packer to load all of my color scheme stuff before loading bufferline (I thought I have it set up to do so but it appears it's not), but checking out this branch works! In case anyone could see anything obvious, here's my config: https://github.com/avegancafe/Juliet/blob/master/nvim/lua/plugins.lua#L192-L198

will:
Thanks this also fixed my issues with rose-pine!

PR #533, opened 2022-09-02
feat: add option to disable name truncation

### LINKED ISSUE #532: [Feature Request]: Auto adjust tab size and auto rename tab filename
### What?

I worked on some projects for most of the files in the projects have very long file names, e.g.

`international_address_suggestion_table.js`, `international_address_suggestion_table.spec.js`,  `international_address_mapping.js`,   `international_address_mapping.spec.js`, 

I have a tab size fixed to 18. But it is hard to tell which file I am working on if I open the above files.
I can reset the tab size to 32, but seems to waste lots of space.
There is an auto-resize feature in barbar.nvim. I think I would need something like that.

Also if it possible to condense the test file name?
e.g. If `international_address_mapping.js`,   `international_address_mapping.spec.js` are shown side by side,
we show 
| international_address_mapping.js | .spec.js|
Or something a-like. 

### Why?

It helps to improve ux in cases when multiple long file names looks similar.

akinsho:
This fixes #532 by allowing a user to set `autosize = true` in `config.options`. With this enabled, users can allow filename to extend up to their full name length.

image

Can you test this out @ray-x, I might change the option name in future commits since I'm not sure I like `autosize` it's really not very descriptive.

### TODO
- [x] Docs
- [x] Tests

### DIFF
--- a/doc/bufferline.txt
+++ b/doc/bufferline.txt
@@ -90,6 +90,7 @@ The available configuration are:
             end,
             max_name_length = 18,
             max_prefix_length = 15, -- prefix used when a buffer is de-duplicated
+            truncate_names = true -- whether or not tab names should be truncated
             tab_size = 18,
             diagnostics = false | "nvim_lsp" | "coc",
             diagnostics_update_in_insert = false,
--- a/lua/bufferline/config.lua
+++ b/lua/bufferline/config.lua
@@ -51,6 +51,7 @@ local constants = lazy.require("bufferline.constants")
 ---@field public separator_style string
 ---@field public name_formatter (fun(path: string):string)?
 ---@field public tab_size number
+---@field public autosize boolean
 ---@field public max_name_length number
 ---@field public color_icons boolean
 ---@field public show_buffer_icons boolean
--- a/lua/bufferline/config.lua
+++ b/lua/bufferline/config.lua
@@ -51,7 +51,7 @@ local constants = lazy.require("bufferline.constants")
 ---@field public separator_style string
 ---@field public name_formatter (fun(path: string):string)?
 ---@field public tab_size number
----@field public autosize boolean
+---@field public truncate_names boolean
 ---@field public max_name_length number
 ---@field public color_icons boolean
 ---@field public show_buffer_icons boolean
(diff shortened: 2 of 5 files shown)


ray-x:
Thanks for the quick PR. It is great and autosize works for me.
However, I saw a issue. It may not related to this PR.  I have 3 file in buffer. However if the windows size is small. There is a corner cases I failed to move to next tab. Please refer to the video. 


https://user-images.githubusercontent.com/1681295/188127433-5d205907-4ccd-4f12-84e6-61b68e832875.mp4

akinsho:
@ray-x I don't think I still understand what is going wrong there. What is the behaviour you are expecting? Just to clarify, the truncation markers are not buttons, so you cannot click on them.

akinsho:
Anyway, if you can clarify the bug feel free to raise an issue. In the meantime I will merge this PR into the `dev` branch for release into the main branch later on.

ray-x:
There are 2 cases that not smoothly. 
* case 1
suppose I have three files, long_file_name_file_1, long_file_name_file_2, long_file_name_file_3. What I saw is
       |  empty_space |long_file_name_file_2|  empty_space  |

There is no way I can use mouse to move to the next buffer.  I have to use `bnext/bprev` (also BufferLinePick not working as only one tab is shown)

* case 2
3 files, as above, but 2 of them are shown in bufferline tab
       |  long_file_name_file_1 |long_file_name_file_2|  
What I expected is if I change focus to long_file_name_file_2, the tab should look like this:
       |  long_file_name_file_2 |long_file_name_file_3|  
If I click on long_file_name_file_2, it should look like this:
       |  long_file_name_file_1 |long_file_name_file_2|  

Is it possible to show at least 2 tabs and the behavior this way (suppose I have 5 files and 3 of them are shown)
       |  long_file... |long_file_name_file_3|  long_file...  |
If I click on `long_file... ` on the right, it will be 
       |  long_file... |long_file_name_file_4|  long_file...  |
If I click on `long_file... ` on the left, it will be 
       |  long_file... |long_file_name_file_2|  long_file...  |

Issue #149, opened 2021-07-18
MRU feature?


mgutz:
Perhaps there's a way to do this but my neovim lua fu isn't as good yours. Is there a way to navigate through the most recently used buffers? I often times will live_grep in telescope, open a few buffers, then I want to quickly get back to the most recently used buffers.

akinsho:
@mgutz nope, there currently isn't an easy way to do this. You can have a look in the docs for a way to sort buffers based on the last modified timestamp which is close-ish to this but if the idea is to do it based on what was last visited then no no exact solution. It's not a trivial problem to solve since nvim doesn't really expose a list or field for what buffers where last opened, it would require building in a mechanism for tracking buffer visits but in a way where the list wasn't constantly shifting in a disorienting way.

I started over in #97 but have abandoned it for now since there were a few bugs and I don't really have time to look into it for the time being. Anyone is free to try to take that over or open a new PR for it though 🤷🏿‍♂️

akinsho:
@mgutz so it turns out that nvim does have a field you can use to see what was last used, buried deep inside a helper method called `getbufinfo` which has a last used field. I found that this is what telescope uses for the buffer picker so I've raised a pr to use that over in #151 can you try that out and see what you think.

Just to be clear it might not be perfect I don't know what mechanism vim is using but as you can tell from the pr it's about 5 lines of code to add, doing it manually will involve a much bigger change so as long as it is "good enough" that's what I'll use. A solution like `telescope-frecency` or something is a huge chunk of work so don't think I'll aim for something like that for now.

mgutz:
I have a naive MRU algorithm working but it's a bit wonky on the UI side. MRU, by its nature changes changes the order of buffers on-the-fly. So, when I cycle to the previous buffer, a buffer might have increased in weight giving it a higher priority than what is visually displayed (as it should). The issue is weight is on BufLeave event based on how long it's been viewed. If a user has viewed a file for an arbitrary duration period, then it becomes the MRU. If a user is just switching through tabs quickly, those tabs have not been viewed long enough to warrant a reordering of the MRU stack.

I think the lastused field is in seconds. I use `reltime` for nanosecond granularity (on linux?) and track each buffer's access. The lastused field may be good for some users but I want something more robust. I'm thinking of giving higher weights to modified files, perhaps auto closing files that have reach negative weight based on time decay.

akinsho:
[...]

PR #535, opened 2022-09-02
Add Sort by MRU #149

bluexnine:
This patch adds sorting by mru feature. The bufferline will sort base
on the last used time. Also add a command to sort by mru.

### DIFF
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -218,6 +218,7 @@ local function setup_commands()
   cmd("BufferLineMovePrev", function() M.move(-1) end, {})
   cmd("BufferLineSortByExtension", function() M.sort_buffers_by("extension") end, {})
   cmd("BufferLineSortByDirectory", function() M.sort_buffers_by("directory") end, {})
+  cmd("BufferLineSortByMru", function() M.sort_buffers_by("mru") end, {})
   cmd(
     "BufferLineSortByRelativeDirectory",
     function() M.sort_buffers_by("relative_directory") end,
--- a/lua/bufferline/sorters.lua
+++ b/lua/bufferline/sorters.lua
@@ -43,6 +43,10 @@ local function sort_by_id(buf_a, buf_b)
   return buf_a.id < buf_b.id
 end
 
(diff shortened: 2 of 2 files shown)


akinsho:
@bluexnine I believe I've tried this strategy before, and it doesn't work very well since the `lastused` variables changes immediately which then means that the position of buffers in this sort is *constantly* changing, e.g. if I moved from left to right the positions of the buffers would likely switch constantly. Ideally an MRU sort would be a little more stable so that it only shifted buffers after a certain amount of time and based on a weighting e.g. it wouldn't reorder things just because you went in and out one time but based on if you were spending large amounts of time in one buffer vs another

bluexnine:
@akinsho You are correct that the buffers does re-organize every time a new buffer is selected. That constant re-org feature is what I wanted and believes it is true to the "most recently used" feature. I think other users may find acceptable. Over time when using this feature, the mru would result in showing the most often buffers first which will reflect as "weighted" since buffers that are not re-select will fall further back and more commonly used buffer will go to the front. 

I do believe weight of time is a feature that sounds interesting. It would need some sort of caching to record use time. The caching and recalculating of used time will make the plugin heavy as more buffers are added. Also another issue I can foresee is that new buffers might always be on the bottom and may confuse users when the new tab does not show up. How do you see new bufffers being organize or do you feel this is not a concern? This feature could be another sort_by feature.

Let me know if you want to adopt the patch. I can always add it as a function.

akinsho:
@bluexnine so the many complexities you raise are exactly why this feature hasn't been implemented yet/why I've considered it as something for a supporting plugin, maybe not sure which yet. If you read through #149 you'll see there's been more discussion around a more complex solution. Personally, I think a solution where the buffers constantly move isn't really user-friendly or at least will lead to a lot of complaints, I'm not ready for.

bluexnine:
@akinsho Thank you for your comments and thoughts. I have read through the issues and they seem to touch upon the same concepts we have touch on. It seems that the definition of the feature you have in mind has not been clearly defined. Let me see if I can ask the census through from the #149 thread and define what you and the users are looking for.

PR #537, opened 2022-09-06
fix: custom area background highlight

jessarcher:
Hi there!

This PR fixes a minor bug introduced with https://github.com/akinsho/bufferline.nvim/pull/497 that prevents the default fill background from being used in custom areas.

I've also updated the references in the test, but I couldn't figure out how to run the tests, so I hope it's okay :woman_shrugging:

Thanks for the fantastic plugin!


akinsho (approved):
LGTM 💯 thanks for the PR

PR #539, opened 2022-09-06
feat(hover): add ability to respond to hover events

akinsho:
### TODO:
- [ ] Tests? fake mouse out and over events..?
- [ ] Hide close button on mouse out

### DIFF
--- a/lua/bufferline/hover.lua
+++ b/lua/bufferline/hover.lua
@@ -1,0 +1,34 @@
+---@diagnostic disable: param-type-mismatch
+local fn, api, map = vim.fn, vim.api, vim.keymap.set
+
+local hover_time = 500
+local hover_timer = nil
+local previous_pos = nil
+
+local M = {}
+
+local function on_hover(current, previous)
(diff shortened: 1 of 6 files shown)


akinsho:
https://user-images.githubusercontent.com/22454918/188915672-fa665d4d-de7b-4129-aace-abdac6393c66.mov

lua/bufferline/hover.lua
```diff
@@ -0,0 +1,55 @@
...
+local version = vim.version()
+
+local delay = 500
+local timer = nil
+local previous_pos = nil
+
+local M = {}
+
+local function on_hover(current)
+  if vim.o.laststatus == 0 then return end
```
vimpostor, line 13: Is there any reason why this is disabled for `laststatus=0`?

This breaks this feature when used with some other plugins (see https://github.com/vimpostor/vim-tpipeline/issues/39), and I don't understand the reasoning behind this check. After all the bufferline still works just fine, if `laststatus=0`. Can we remove this check?

PR #548, opened 2022-09-14
Pass more information to name_formatter

rwblokzijl:
With this PR, the following fields will be passed to `name_formatter`

| Field            | Type   | Description                                   |
| ---------------- | ------ | --------------------------------------------- |
| name             | str    | the basename of the active file               |
| path             | str    | the full path of the active file              |
| bufnr            | int    | the number of the active buffer               |
| extension        | str    | the file extension of the active file         |
| icon             | str    | the icon of the active file                   |
| icon_highlight   | str    | the highlight group of the active file        |
| ordinal          | int    | the ordinal number of the buffer/tab          |
| modified         | bool   | whether the active file has unwritten changes |
| modifiable       | bool   | whether the active file is modifiable         |

for tabs the following fiels are also included:

| Field       | Type         | Description                             |
| ----------- | ------------ | --------------------------------------- |
| tabnr       | int          | the number of the tab                   |
| focusable   | bool         | whether the tab is focusable            |
| buffers     | table(int)   | the numbers of the buffers in the tab   |
| hidden      | bool         | whether the tab is hidden               |

### DIFF
--- a/doc/bufferline.txt
+++ b/doc/bufferline.txt
@@ -83,11 +83,22 @@ The available configuration are:
             --- Please note some names can/will break the
             --- bufferline so use this at your discretion knowing that it has
             --- some limitations that will *NOT* be fixed.
-            name_formatter = function(buf)  -- buf contains a "name", "path" and "bufnr"
-                -- remove extension from markdown files for example
-                if buf.name:match('%.md') then
-                    return vim.fn.fnamemodify(buf.name, ':t:r')
-                end
+            name_formatter = function(buf)  -- buf contains:
+                  -- name           | str  | the basename of the active file
+                  -- path           | str  | the full path of the active file
+                  -- bufnr          | int  | the number of the active buffer
+                  -- extension      | str  | the file extension of the active file
+                  -- icon           | str  | the icon of the active file
+                  -- icon_highlight | str  | the highlight group of the active file
+                  -- ordinal        | int  | the ordinal number of the buffer/tab
+                  -- modified       | bool | whether the active file has unwritten changes
+                  -- modifiable     | bool | whether the active file is modifiable
+
+                  -- for tabs the following fiels are also included:
+                  -- tabnr          | int        | the number of the tab
+                  -- focusable      | bool       | whether the tab is focusable
+                  -- buffers        | table(int) | the numbers of the buffers in the tab
+                  -- hidden         | bool       | whether the tab is hidden
             end,
             max_name_length = 18,
             max_prefix_length = 15, -- prefix used when a buffer is de-duplicated
--- a/lua/bufferline/models.lua
+++ b/lua/bufferline/models.lua
@@ -139,7 +139,22 @@ function Tabpage:new(tab)
     type = tab.buftype,
   })
   if tab.name_formatter and type(tab.name_formatter) == "function" then
-    tab.name = tab.name_formatter({ name = tab.name, path = tab.path, tabnr = tab.id }) or tab.name
+    tab.name = tab.name_formatter({
+      name           = tab.name,
+      path           = tab.path,
+      bufnr          = tab.buf,
+      extension      = tab.extension,
+      icon           = tab.icon,
+      icon_highlight = tab.icon_highlight,
+      ordinal        = tab.ordinal,
+      modified       = tab.modified,
+      modifiable     = tab.modifiable,
+
+      tabnr          = tab.id,
+      focusable      = tab.focusable,
+      buffers        = tab.buffers,
+      hidden         = tab.hidden,
+      }) or tab.name
   end
   setmetatable(tab, self)
   self.__index = self
@@ -221,7 +236,17 @@ function Buffer:new(buf)
     name = fn.fnamemodify(buf.path, ":t")
     name = is_directory and name .. "/" or name
     if buf.name_formatter and type(buf.name_formatter) == "function" then
-      name = buf.name_formatter({ name = name, path = buf.path, bufnr = buf.id }) or name
+      name = buf.name_formatter({
+        name = name,
+        path = buf.path,
+        bufnr = buf.id,
+        extension = buf.extension,
+        icon = buf.icon,
+        icon_highlight = buf.icon_highlight,
+        ordinal = buf.ordinal,
+        modified = buf.modified,
+        modifiable  = buf.modifiable
+      }) or name
     end
   end
   buf.name, buf.filename = name, name -- TODO: remove this 'filename' field


akinsho:
@rwblokzijl at first glance I can only guess that the intended use case for all this information is to actually use the name formatter as a way of custom rendering the buffer info i.e. putting in symbols for different states etc.

The name formatter field was always intend specifically for change the name only, not as a backdoor for custom rendering the buffer tab. I have ideas for how that can be implemented and have been slowly refactoring the plugin with that eventual goal in mind.

To my mind, this seems like *waaay* too much information to pass to the name formatter. It is essentially the entire contents of a buffer class or a tab, and I don't currently understand why all this information is necessary, other than as a backdoor to rendering different things than just the name.

Can you describe the specific problem or use case you are trying to solve for? Passing any data to the user forces my hand for future refactorings as people will come to depend on fields that I might rename, remove or refactor completely. This will increase or leak a lot of implementation/state implicitly, which is something I would only do having thought long and hard about it.

akinsho:
For more context the name formatter was added *reluctantly* as a user had one very specific use case to use this with an external plugin. It isn't intended as the entry point for custom rendering (just re-iterating this point)

rwblokzijl:
I picked bufferline specifically for this feature. My workflow uses 1 tab per module, and i wanted to show the module name in the tab instead of the filename. (In terraform the filename is always "main.tf"). So i compute the parent folder of all the paths of the buffers in the tab. This meant: 
1. Given the tabnr, get the windows for that tab
2. Given those windows, get the buffers for those windows
3. Given those buffers get their paths
4. Given those paths compute the module name

During this i ran into an issue: The `tabnr` provided was not the tab number vim used, but some internal bufferline `id` (i needed `ordinal`).

While adding `ordinal` to the code i thought: why not add all the information. Specifically `buffers` ended up being useful for me, since it allowed me to skip step 1 and 2.

---

Everything above is my actual argument, if you only want to add those two fields i would also be very happy. However, in ADDITION to generating a custom name, i ALSO use the name_formatter to do *some* formatting.

I use a powerline style lualine and wanted to make bufferline look similar, so i did something VERY dirty and hacky. I made my bufferline look like this:
2022-09-14-161343_947x58_scrot

<details>
  <summary>My config</summary>
    
``` lua
require("bufferline").setup {
  options = {
    mode = "tabs",
    numbers = "ordinal",
    diagnostics = "nvim_lsp",
    -- -- The diagnostics indicator can be set to nil to keep the buffer name highlight but delete the highlighting
    diagnostics_indicator = function(count, level)
      local icons = {
        error   = " ",
        warning = " ",
        info    = " ",
        hint    = "",
      }
      return icons[level] .. count
    end,
    -- separator_style = {"right_of_active", "right_of_inactive"} --     
    separator_style = {"", ""},
    indicator = {
      -- style = "icon",
      icon = "", -- Effectively the left seperator for the active tab
    },
    left_trunc_marker = '',
    right_trunc_marker = '',

    modified_icon = '█ ', buffer_close_icon = '█ ',
    close_icon = 'CLOSE',
    show_close_icon = false,
    -- show_buffer_close_icons = false,
    tab_size = 28,
    name_formatter = function(tab)  -- tab contains a "name", "path" and "bufnr"
      local bufs = {}
      for _, buf_nr in ipairs(tab.buffers) do
        local buf_info = vim.fn.getbufinfo(buf_nr)[1]
        table.insert(bufs, buf_info)
      end
      local modified = false
      for _, buf_info in ipairs(bufs) do
        if buf_info.changed == 1 then
          modified = true
        end
      end
      local name = ""
      for _, buf_info in ipairs(bufs) do
        if buf_info.name:match(".*NvimTree_%d*$") then -- Dont change name for nvimtree
          goto continue
        elseif buf_info.name:startswith("term://") then -- Dont change name for fzf
          goto continue
        elseif buf_info.name:match('%.tf') then
          name = buf_info.name:path_parent()
          break
        else
          name = buf_info.name:basename() or tab.name
          break
        end
        ::continue::
      end
      return name .. (modified and ' [+]' or '')
    end,
    offsets = {
      {filetype = "NvimTree", text = "0. פּ File Explorer" , text_align = "left", highlight="lualine_b_normal"},
    },
    -- color_icons = true | false, -- whether or not to add the filetype icon highlights
    -- show_buffer_icons = true | false, -- disable filetype icons for buffers
    -- show_buffer_default_icon = true | false, -- whether or not an unrecognised filetype should show a default icon
    -- show_tab_indicators = true,
    -- persist_buffer_sort = true, -- whether or not custom sorted buffers should persist
    enforce_regular_tabs = true,
  },

  -- :h bufferline-styling
  highlights = {
    modified              = colors.getTransisionColors("lualine_b_normal", "lualine_c_inactive"),
    modified_visible      = red,
    modified_selected     = colors.getTransisionColors("lualine_a_normal", "lualine_c_inactive"),

    close_button          = colors.getTransisionColors("lualine_b_normal", "lualine_c_inactive"),
    close_button_visible  = red,
    close_button_selected = colors.getTransisionColors("lualine_a_normal", "lualine_c_inactive"),

    separator             = colors.getTransisionColors("lualine_c_inactive", "lualine_b_normal"),
    separator_visible     = red,
    separator_selected    = colors.getTransisionColors("lualine_a_normal", "lualine_c_inactive"),
    indicator_selected    = colors.getTransisionColors("lualine_b_normal", "lualine_a_normal"),

    numbers               = colors.getColorComponents("lualine_b_normal"),
    numbers_visible       = red,
    numbers_selected      = colors.getColorComponents("lualine_a_normal"),

    background            = colors.getColorComponents("lualine_b_normal"),
    fill                  = colors.getColorComponents("lualine_c_inactive"),

    buffer                = red,
    buffer_visible        = red,
    buffer_selected       = colors.getColorComponents("lualine_a_normal"),

    -- diagnostic                  =
    -- diagnostic_visible          = red,
    -- diagnostic_selected         =

    hint                        = colors.getColorComponents("lualine_b_normal"),
    hint_visible                = red,
    hint_selected               = colors.getColorComponents("lualine_a_normal"),
    hint_diagnostic             = colors.getCombinedColors("DiagnosticHint", "lualine_b_normal"),
    hint_diagnostic_visible     = red,
    hint_diagnostic_selected    = colors.getColorComponents("lualine_a_normal"),
    info                        = colors.getColorComponents("lualine_b_normal"),
    info_visible                = red,
    info_selected               = colors.getColorComponents("lualine_a_normal"),
    info_diagnostic             = colors.getCombinedColors("DiagnosticInfo", "lualine_b_normal"),
    info_diagnostic_visible     = red,
    info_diagnostic_selected    = colors.getColorComponents("lualine_a_normal"),
    warning                     = colors.getColorComponents("lualine_b_normal"),
    warning_visible             = red,
    warning_selected            = colors.getColorComponents("lualine_a_normal"),
    warning_diagnostic          = colors.getCombinedColors("DiagnosticWarn", "lualine_b_normal"),
    warning_diagnostic_visible  = red,
    warning_diagnostic_selected = colors.getColorComponents("lualine_a_normal"),
    error                       = colors.getColorComponents("lualine_b_normal"),
    error_visible               = red,
    error_selected              = colors.getColorComponents("lualine_a_normal"),
    error_diagnostic            = colors.getCombinedColors("DiagnosticError", "lualine_b_normal"),
    error_diagnostic_visible    = red,
    error_diagnostic_selected   = colors.getColorComponents("lualine_a_normal"),

    tab                   = red,
    tab_selected          = red,
    tab_close             = red,

    duplicate                   = red,
    duplicate_visible           = red,
    duplicate_selected          = red,

    pick                        = red,
    pick_visible                = red,
    pick_selected               = red,
  }
}
```
</details>

I misused the modified symbol to make my bar look prettier, but it doesn't work when the tab isn't selected. So i also misused the `name_formatter` to make re-add a motified symbol.

I expect NO support for this hacky workaround, however more information in the `name_formatter` would have made my initial implementation of the module naming easier, and including the `modified` would have made my hacky ricing a lot easier.

akinsho:
@rwblokzijl thanks for clarifying. I also work with terraform, so the reasoning makes sense. So I'm fine to add the buffers within a tab as well as the tab number to the tab name formatter.

> The tabnr provided was not the tab number vim used, but some internal bufferline id (I needed ordinal).

This sounds like a bug, the ordinal is very much an internal thing and just represents the actual order in bufferline's list, but I would consider it quite private since I don't love this variable and definitely don't want it exposed. The `tabnr` should be correct and if it isn't that needs fixing.

Regarding changing styling, I've slowly refactored this plugin so that actually each tab is derived from a list of tables
https://github.com/akinsho/bufferline.nvim/blob/ffa1d0009f5e52d3d32d08c9a781f18b2b2ef507/lua/bufferline/ui.lua#L517-L534

I'm considering exposing this once I have a clear idea of what the right API is to expose this somehow to the user, so they can shuffle things around and make some tweaks. It needs to be done carefully as there's a lot of room for breakage or the wrong API needing a bunch of future breaking changes.

Lastly, I haven't entirely decided if I want to expose this at all. I appreciate that people have wildly different preferences but ultimately the main design/use case for why I built and continue to maintain this plugin is to have something that looks like sublime, atom, vscode etc. tab bar. People looking for that are really my main target audience.

Anyway, that was a long ramble to say I'm fine with exposing the bare minimum fields you described, e.g. `tabnr` and `buffers`

rwblokzijl:
Sounds good, i will update the PR soon.

I also looked a bit further into the oridinal thing, and i found this:
- neovim has the concept of a tab "handle" and the concept of a tab "page"
- a list of tab "handles" might look like {1, 4, 2, 3}, the handles are created in order of their number. If tab 2 is removed it will look like:  {1, 4, 3}, with no tabs changing numbers. 
- a list of tab "pages" with always be ordinal, eg {1, 2, 3}, if tab 2 is removed it will look like {1, 2} with tab 3 becoming the new tab 2. 
- given the tabpage handle, the tabpage number can be gotten with: `vim.api.nvim_tabpage_get_number(handle)`
    - It basically maps to the index in the table. EG {1, 4, 3} -> {1, 2, 3}
- all vimscript calls regarding tabs use the "page". Just like keybinding `1gt`, `4gt` etc.
- bufferline fetches the list of tab "handles" with: `vim.api.nvim_list_tabpages()`
- bufferline passes the "handle" to the `name_formatter` under the label `tabnr`

If i have `buffers` i dont need `tabnr` at all anymore, but if i wanted the actual number i could just convert it with `nvim_tabpage_get_number`.

This is all very unclear to users getting the "handle" so i propose i add `buffers` to the inputs to `name_formatter` and add a reference to `vim.api.nvim_tabpage_get_number` in the docs.

rwblokzijl:
I opted not to change `tabnr` to a more descriptive `tabhandle` for backwards compatibility.

akinsho:
Ah, that makes sense, I thought it would be something like that, similar to the whole window number vs ID thing. This LGTM now 👍🏿

akinsho:
Just merged this in. Please note I've merged it into the `dev` branch which is where all changes go now until there's a release

PR #554, opened 2022-09-16
refactor(tabs): tab is modified if any child buffer is modified

rwblokzijl:
If a buffer is modified within a tab, the tab is only marked as modified if that buffer is in focus. This PR marks the tab as modified if any of the buffers in the tab are modified.

### DIFF
--- a/lua/bufferline/models.lua
+++ b/lua/bufferline/models.lua
@@ -129,7 +129,10 @@ function Tabpage:new(tab)
   tab.name = fn.fnamemodify(tab.path, ":t")
   assert(tab.buf, fmt("A tab must a have a buffer: %s", vim.inspect(tab)))
   tab.modifiable = vim.bo[tab.buf].modifiable
-  tab.modified = vim.bo[tab.buf].modified
+  tab.modified = false
+  for _, buf in pairs(tab.buffers) do
+    tab.modified = tab.modified or vim.bo[buf].modified
+  end
   tab.buftype = vim.bo[tab.buf].buftype
   tab.extension = fn.fnamemodify(tab.path, ":e")
   tab.icon, tab.icon_highlight = utils.get_icon({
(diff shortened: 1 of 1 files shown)


lua/bufferline/models.lua
```diff
@@ -129,7 +129,10 @@ function Tabpage:new(tab)
   tab.name = fn.fnamemodify(tab.path, ":t")
   assert(tab.buf, fmt("A tab must a have a buffer: %s", vim.inspect(tab)))
   tab.modifiable = vim.bo[tab.buf].modifiable
-  tab.modified = vim.bo[tab.buf].modified
+  tab.modified = false
+  for _, buf in pairs(tab.buffers) do
```
akinsho, line 133: Can this be factored out, I'd rather avoid logic being done inside this function directly and have it be more just set statements like `tab.modified = get_modified_state(buffers)`

akinsho (changes requested):
This LGTM other than one comment I left. It's worth noting that whilst this seems fine to me, I can't predict whether some other tabs user will turn around and complain that they prefer the alternate way of working 🤷🏿‍♂️. Might call on you to defend this but definitely would rather only one mechanism here and this seems sensible

lua/bufferline/models.lua
rwblokzijl: Yeah, I'll do that. Should i leave the function in the same file?

akinsho: Yes, it makes the most sense in that file. It's really just a very tiny stylistic preference. I just want to be able to look at that function and see each value being set directly somehow rather than following any logic around

rwblokzijl:
Sounds good. I cannot personally think of a workflow where this feature would be unwanted, but if one comes along I'll join the discussion.

akinsho (approved):
LGTM 🚀 thanks for tweaking it 👍🏿

Issue #469, opened 2022-07-06
[Feature Request]: Toggle unique buffer name


CharaD7:
### What?

It would be great to be able to toggle on and off unique buffer name.

### Why?

I tried to show unique buffer name but could not find a way to do this. However, I think sometimes. we want to see unique buffer names, other times, we don't so it won't hurt to have a feature to allow one to easily toggle unique buffer name.

akinsho:
@CharaD7 I'm not against this, so if you'd like to raise a PR for it, should quite straightforward. You will need to add (yet another) option to the `config.lua` which would default to `true` then. Where the duplicate is added you will need to add a check for that value and if it is empty don't add the duplicate. This should be in `buffers.lua` and `tabpages.lua`.

I don't really see this as critical since after almost 2 years this is the first time anyone has requested this, so I'm probably not going to do this myself anytime soon but maybe in the distant future. If you want it, it will be best for **you** to raise a PR.

yyy33:
> @CharaD7 I'm not against this, so if you'd like to raise a PR for it, should quite straightforward. You will need to add (yet another) option to the `config.lua` which would default to `true` then. Where the duplicate is added you will need to add a check for that value and if it is empty don't add the duplicate. This should be in `buffers.lua` and `tabpages.lua`.
> 
> I don't really see this as critical since after almost 2 years this is the first time anyone has requested this, so I'm probably not going to do this myself anytime soon but maybe in the distant future. If you want it, it will be best for **you** to raise a PR.

For me this is really useful, I am trying to replace tmux with neovim as a terminal, however the name of each terminal is the name of a shell, like zsh, so when I open as much as a terminal the bufferline gets ugly, so I would like to disable this feature when the buffer type is terminal, I tried to read the source code but it is a bit hard because my English is not very good, can you tell me a temporary solution?

Translated with www.DeepL.com/Translator (free version)

![2022-07-31_19-04](https://user-images.githubusercontent.com/51548147/182023436-81300903-1104-4147-8c05-670ab3249880.png)
![2022-07-31_19-04_1](https://user-images.githubusercontent.com/51548147/182023445-a2378b05-4e0f-4a8f-8072-67ee9652a61b.png)

akinsho:
@CharaD7 the comment I made earlier still holds, PRs welcome 🙏🏿

akinsho:
This is now fixed by #556 in `dev` and will be released soon

PR #556, opened 2022-09-18
feat(duplicates): toggle unique buffer name (#469)

shans10:
Added a boolean option to enable or disable duplicate buffer prefix.


akinsho (approved):
LGTM 👍🏿 thanks for the PR

PR #558, opened 2022-09-18
refactor(tabs): diagnostics are the sum of buffer diagnostics

rwblokzijl:
In the same spirit as https://github.com/akinsho/bufferline.nvim/pull/554. This PR combines the diagnostic information shown on the line.

If the same file is open multiple times in the same tab, it is only included once.

### DIFF
--- a/lua/bufferline/diagnostics.lua
+++ b/lua/bufferline/diagnostics.lua
@@ -19,6 +19,15 @@ local severity_name = {
   [5] = "other",
 }
 
+local function get_severity_code(sev_name)
+  for code, name in pairs(severity_name) do
+    if name == sev_name then
+      return code
+    end
+  end
+  return 5
+end
+
 setmetatable(severity_name, {
   __index = function() return "other" end,
 })
@@ -113,6 +122,27 @@ local get_diagnostics = {
     return function() return diagnostics end
   end)(),
 }
+
+function M.combine(diagnosticss)
+  if #diagnosticss < 1 then return {} end
+  local result = {
+    sev_code = get_severity_code(diagnosticss[1].level),
+    errors = {},
+    count = 0
+  }
+  for _, diagnostics in pairs(diagnosticss) do
+    result.sev_code = math.min(result.sev_code, get_severity_code(diagnostics.level))
+    for severity, count in pairs(diagnostics.errors) do
+      result.errors[severity] = count + (result.errors[severity] or 0)
+      result.count = count + result.count
+    end
+  end
+  return {
+      level = severity_name[result.sev_code],
+      errors = result.errors,
+      count = result.count
+    }
+end
 
 ---@param opts table
 function M.get(opts)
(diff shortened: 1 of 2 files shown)


lua/bufferline/diagnostics.lua
```diff
@@ -19,6 +19,15 @@ local severity_name = {
   [5] = "other",
 }
 
+local function get_severity_code(sev_name)
```
akinsho, line 22: I think this can be simplified/changed to use `vim.tbl_add_reverse_lookup` so you could then delete this entire function and instead do
```lua
local severity_name = vim.tbl_add_reverse_lookup({
  [1] = "error",
  [2] = "warning",
  [3] = "info",
  [4] = "hint",
  [5] = "other",
}) -- this will add {error = 1, warning = 2, info = 3, ...}

-- Place where you need the code
severity_name[severity]
```

akinsho (changes requested):
@rwblokzijl thanks for the PR 🚀, left a few comments about tweaking some functions, using builtins where possible. Otherwise looks good 👍🏿

lua/bufferline/tabpages.lua
```diff
@@ -98,6 +98,29 @@ local function get_tab_buffers(tab_num)
   return vim.tbl_map(api.nvim_win_get_buf, api.nvim_tabpage_list_wins(tab_num))
 end
 
+local function contains_value(table, element)
```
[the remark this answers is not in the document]
rwblokzijl: I didn't know about `vim.tbl_contains` I'll change it :)

lua/bufferline/diagnostics.lua
```diff
@@ -114,6 +123,27 @@ local get_diagnostics = {
   end)(),
 }
 
+function M.combine(diagnosticss)
```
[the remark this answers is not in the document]
akinsho: @rwblokzijl it's more that the word `diagnosticss` is a typo in English. The word `diagnostic` is singular and `diagnostics` is plural. In your case you are adding two `ss` which is not correct.

lua/bufferline/diagnostics.lua
```diff
@@ -114,16 +114,16 @@ local get_diagnostics = {
...
   local result = {
-    sev_code = severity_name[diagnosticss[1].level],
+    sev_code = severity_name[diagnostics[1].level],
     errors = {},
     count = 0
   }
-  for _, diagnostics in pairs(diagnosticss) do
-    result.sev_code = math.min(result.sev_code, severity_name[diagnostics.level])
-    for severity, count in pairs(diagnostics.errors) do
+  for _, diagnostics_instance in pairs(diagnostics) do
```
akinsho, line 124: This should just use the singluar here which is
```suggestion
  for _, diagnostic in pairs(diagnostics) do
```

rwblokzijl: Yes, but that would reuse the variable name.

akinsho: In which case something like `d` or `diag` is fine. I think it's fine to be terse here since it's to my mind pretty clear what each individual item is

akinsho (approved):
LGTM thanks for making the tweaks 👍🏿

PR #563, opened 2022-09-22
feat: let commands "BufferLineCloseRight" and "BufferLineCloseRight" skip unwritten buffers without a bang

yangyuanxx:
This PR changes the default behavior of commands "BufferLineCloseRight" and "BufferLineCloseRight": unwritten buffers will be skipped unless a bang is added.


lua/bufferline/commands.lua
```diff
@@ -46,10 +46,14 @@ local function open_element(id)
 end
 
 ---@param id number
-local function delete_element(id)
+local function delete_element(id, force)
   if config:is_tabline() then
     vim.cmd("tabclose " .. id)
   else
+		local is_modified = api.nvim_buf_get_option(id, 'modified')
+		if is_modified and not force then
```
akinsho, line 54: The indentation here is off, can you align this with other code at this level e.g line 51. Also
```suggestion
if vim.bo[id].modified and not force then
```

lua/bufferline/models.lua
```diff
@@ -127,9 +127,7 @@ local Tabpage = Component:new({ type = "tab" })
 
 local function get_modified_state(buffers)
   for _, buf in pairs(buffers) do
-    if vim.bo[buf].modified then
```
akinsho, line 130: I think you need to *rebase* your branch not merge in the dev branch since this just adds noise to your PR and makes it look like you added in these unrelated changes

tests/bufferline_spec.lua
```diff
@@ -223,7 +223,26 @@ describe("Bufferline tests:", function()
       bufs = vim.api.nvim_list_bufs()
       assert.is_equal(1, #bufs)
     end)
+
+    it("should close buffers in direction, but skip unwritten ones", function()
```
akinsho, line 227: Thanks for adding a test

akinsho (changes requested):
@yangyuanxx thanks for raising this PR. It's much appreciated, I've left a few stylistic comments and recommend you rebase your branch to remove the duplicated commits from my dev branch. Otherwise, this LGTM

lua/bufferline/commands.lua
akinsho: @yangyuanxx I don't think you saw the review comment about changing the `is_modified` variable to just use the `vim.bo[id]` syntax

lua/bufferline/models.lua
akinsho: @yangyuanxx can you rebase this branch on dev it has a lot of unrelated commits as part of it

akinsho (commented):
@yangyuanxx there are still pending review comments

akinsho:
@yangyuanxx thanks for the PR, I've merged it in locally and made a few tweaks to use the default behaviour of `nvim_buf_delete` i.e. rather than manually checking for a modified state and reproducing vim's behaviour just pass the `force` argument directly to `nvim_buf_delete` instead.

The commit is now in dev 78a65d2d207553b0ce3d1e524e3f4e65a752cda4

PR #571, opened 2022-09-28
feat: reuse "close_command" to implement element deleting.

yangyuanxx:
This PR modifies the behavior of BufferLineClose* commands which will depends on the "close_command" config.
The main motivation is to ensure consistent behavior in different use cases of deleting buffers.


akinsho:
@yangyuanxx thanks for the PR, one thing that concerns me about it is that this might end up being a breaking change for some users. Given that the user is completely free to handle how the `close_command` works, they could easily have put a function there that is not compatible with closing a bunch of buffers in series 🤔

yangyuanxx:
@akinsho Thanks for reply. I fully understood your concern. 
In my opinion, it is something about how to define the "close_command"'s responsibility: generic or limited.
Generic leads to more consistent behavior, and limited leads to lower coupling. And for this feature, I may prefer the former.

If we end up making this change, we should note this breaking change in the migration guide of version 3.

akinsho:
@yangyuanxx in which case can you point this PR towards the v3.0.0 branch if we were to merge it in that is where it should go

akinsho:
Actually I need to rebase that branch first

yangyuanxx:
@akinsho  Sorry for the late reply. I created a new PR on branch v3.0.0: #572

PR #572, opened 2022-09-30
feat: reuse "close_command" to implement element deleting

yangyuanxx:
As discussed here: #571


akinsho:
@yangyuanxx would you be able to create this as a *separate branch* into `v3.0.0` that way I can merge it into that branch. As it is the commits are directly on the `v3.0.0` branch and this branch is now out of date

PR #573, opened 2022-10-01
fix(custom_area): default = false for highlights

lambdahands:
Hello!

I noticed an issue in which the highlights in my dynamic custom areas were not updating conditionally while the text contents changed as expected. I dug deeper, and it seems that `default = true` is being passed to `vim.api.nvim_set_hl`, which does the _opposite_ of what the code comment states as the intention of setting it to `true`:

```lua
-- We need to be able to constantly override these highlights so they should always be default
opts.default = true
```

Changing `opts.default = false` seems to resolve my issue, and highlights update as expected in my custom areas.

Let me know if this needs any further editing or if I went down the wrong path and there's a simpler way to update custom area highlights. Thanks! :)

**Edit:** I noticed a similar issue with setting `config.options.themable`: the highlighting function treats `themable = true` as `default = true`, which means that the `themable` option doesn't allow for highlights to be overwritten after they are set once.

### DIFF
--- a/lua/bufferline/highlights.lua
+++ b/lua/bufferline/highlights.lua
@@ -119,7 +119,7 @@ end
 function M.set_one(name, opts)
   if not opts or vim.tbl_isempty(opts) then return end
   local hl = filter_invalid_keys(opts)
-  hl.default = vim.F.if_nil(opts.default, config.options.themable)
+  hl.default = vim.F.if_nil(opts.default, not config.options.themable)
   local ok, msg = pcall(api.nvim_set_hl, 0, name, hl)
   if ok then return hl end
   utils.notify(
--- a/lua/bufferline/custom_area.lua
+++ b/lua/bufferline/custom_area.lua
@@ -21,7 +21,7 @@ local function create_hl(index, side, section, bg)
   local opts = highlights.translate_user_highlights(section)
   opts.bg = opts.bg or bg
   -- We need to be able to constantly override these highlights so they should always be default
-  opts.default = true
+  opts.default = false
   highlights.set_one(name, opts)
   return highlights.hl(name)
 end


lua/bufferline/highlights.lua
```diff
@@ -119,7 +119,7 @@ end
 function M.set_one(name, opts)
   if not opts or vim.tbl_isempty(opts) then return end
   local hl = filter_invalid_keys(opts)
-  hl.default = vim.F.if_nil(opts.default, config.options.themable)
+  hl.default = vim.F.if_nil(opts.default, not config.options.themable)
```
akinsho, line 122: This change will change highlighting for *all of bufferline's highlights* and is almost certainly not what is intended. Please do not apply the change you desire here.

akinsho (commented):
@lambdahands thanks for the PR, although this isn't a breaking change for most people, changing how every highlight is set is almost certainly the wrong approach.

It's important to note that the issue with highlighting can very often be related to how a user is applying highlights in the first place. This commit might fix things for you, but will very likely break things for most people, myself included.

The decision to apply these highlights is very intentional, please see `:help highlight-default`. The intention here is that if a use sets `themable` then all highlights are `default` which means they are only set if an existing highlight does not already exist. This means that colour schemes and users are able to override bufferline highlights. This is by design, the specific change you are trying to remove/invert came into being due to quite a bit of trial and error.

akinsho:
@lambdahands it would be good to have a clear issue of what problem you are trying to solve since context that you are likely missing is that I added in these changes specifically to fix the issue of highlighting for other users. This change has previously regressed and brought every back to my door with complaints of breakage. I'm not looking to bring that issue up again. By setting highlights as default, this plugin essentially defers highlighting to those setup elsewhere i.e. overrides

lambdahands:
Hi @akinsho, thanks for the reply and the explanation of how this behavior came to be. It seems I was stumbling into an edge case and wound up confused as to the purpose of the `themable` option. I'll close this out and open an issue that describes my specific use-case as I think it is still helpful (I'm essentially adding a "battery" indicator that changes color depending on my laptop charge). Thanks again for going into detail about how the highlighting system works!

PR #576, opened 2022-10-04
feat: reuse "handle_close" in "delete_element" so users could expect more consistent behavior

yangyuanxx:
As mentioned here: https://github.com/akinsho/bufferline.nvim/pull/572


akinsho (approved):
Just one minor tweak then good to go

Issue #580, opened 2022-10-07
[Feature Request]: Commands for "Close all buffers" and "Close All Buffers Except Current"


andy-bell101:
### What?

I'd like some commands to be able to close all the buffers in my workspace, and one to close all but the current buffer.

### Why?

The neovim builtin commands for closing buffers (e.g. `:%bd`) mess too much with my workspace. Bufferline already has commands for closing buffers in a more intelligent way, so it seems like a logical place to add some extra functionality.

andy-bell101:
Exposing some internals in #583 that will allow users to do this themselves

PR #581, opened 2022-10-07
[feat] Add "Close All" and "Close All But Current" commands

andy-bell101:
See #580

### DIFF
--- a/doc/bufferline.txt
+++ b/doc/bufferline.txt
@@ -498,6 +498,9 @@ are:
   current buffer
 * `BufferLineCloseLeft` - close all visible buffers to the left of the current
   buffer
+* BufferLineCloseAll - close all visible buffers
+* BufferLineCloseAllButCurrent - close all visible buffers except the current
+  one
 
 NOTE: these commands will skip unwritten buffers, you can add `!` to close without writing.
 
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -41,6 +41,8 @@ local M = {
   sort_by = commands.sort_by,
   pick_buffer = commands.pick,
   close_with_pick = commands.close_with_pick,
+  close_all = commands.close_all,
+  close_all_but_current = commands.close_all_but_current,
   close_in_direction = commands.close_in_direction,
   -- @deprecate
   go_to_buffer = commands.go_to,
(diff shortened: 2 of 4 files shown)


akinsho:
@andy-bell101 thanks for the PR, we've made some other changes to deletion which now conflict here, unfortunately.

Regarding the PR in general whilst these changes generally look fine, I'd avoided this sort of command for a long time because there are vim native commands that do both of these things and this is a bit like re-inventing the wheel and most importantly it puts this plugin in the position of managing buffers which I really wanted to avoid. In fairness, it's my fault since I added commands like `CloseLeft` and `CloseRight` which I should never have done since that opens the door for people asking about only etc. instead of learning to use vim's native commands e.g.

>                   :.,$-bdelete   "delete buffers from the current one to
>                                  " last but one
>                   :%bdelete      " delete all buffers

From the `:help bdelete` docs there are similar ones for tabs as well.

I'm not against this change, but I worry that the next thing will be "can bufferline also handle <insert buffer management command> please", which is already a request I'm seeing too often. Since this plugin is for UI only really.

andy-bell101:
I should be able to fix the breaking changes some time this weekend (assuming it's worth continuing).

My issue with `%bd` is that it's too agressive if that makes sense? For example, I use `neotree` and `:%bd` will close that too, which I don't want. I'm not sure where exactly it happens, but bufferline seems to filter out the `neotree` buffer which makes the function I wrote do what I want. And I can use the custom filter in the config to filter out other buffers from other plugins that I want (for example, dapui).

I understand where you're coming from though in trying to keep the plugin 'pure'. If you think it's worth carrying on I'll fix the conflicts. If not, I'll just keep it in my own fork.

Thanks for the plugin by the way!

akinsho:
@andy-bell101 that's a good point and the way it works is that bufferline ignores any buffers that are not listed, which most plugin authors will set for their special plugin buffers so it will ignore those by default. Tbh I don't mind adding these since its come up quite often. I think with this addition though, it would be good to also clarify in the README that this plugin is not a buffer manager and no more commands like this will be added.

Another alternative is to provide like an api function to give a user access to bufferline's elements not in full since I wouldn't want to expose the internals but something like.

```lua
require('bufferline').get_elements()
-- which returns
{
  mode = "tabs",
  -- this should be a subset of the actual element as I don't want to expose internal fields
  -- it should also ignore things like group indicators etc. which I think ere is a flag for
  elements = {{id = <num>, name = "example"}}
}
```

With the above it would just be a matter of documenting in the wiki how to implement a close all and close except current and adding these examples to the docs

andy-bell101:
I think your second idea is better. That way bufferline line can stay doing only "bufferline" things but users can operate on buffers themselves if they need to. I'll take a look at implementing your suggestion tonight

andy-bell101:
For simplicity I'm just going to dump this PR because I can't be bothered to fix the git mess this has created

kohane27:
If you want to "Close All But Current", the following function works:

```lua
function vim.close_all_but_current_buffer()
  for _, e in ipairs(require("bufferline").get_elements().elements) do
    vim.schedule(function()
      if e.id == vim.api.nvim_get_current_buf() then
        return
      else
        vim.cmd("bd " .. e.id)
      end
    end)
  end
end
```

The code is from `:h bufferline-working-with-elements`, I just added the condition to check if it's the current buffer.

Cheers!

PR #583, opened 2022-10-09
feat: add `get_elements` command

andy-bell101:
As discussed in #581 

I've included the path to the buffer in the output as well as I thought that would be useful. Otherwise it should be outputting what you were expecting.

There doesn't seem to be a wiki yet so I added the information to the README and vim docs.


lua/bufferline/commands.lua
```diff
@@ -191,6 +191,18 @@ function M.cycle(direction)
   open_element(item.id)
 end
 
+function M.get_elements()
+    -- there must be a better way of doing this but I don't know Lua well
+    local elems = {}
+    for _, e in ipairs(state.components) do
```
akinsho, line 197: You can use `vim.tbl_map` i.e.

```lua
return vim.tbl_map(function(elem)
	return {...}
end, state.components)
```

lua/bufferline.lua
```diff
@@ -228,6 +229,7 @@ local function setup_commands()
   cmd("BufferLinePickClose", function() M.close_buffer_with_pick() end, {})
   cmd("BufferLineCycleNext", function() M.cycle(1) end, {})
   cmd("BufferLineCyclePrev", function() M.cycle(-1) end, {})
+  cmd("BufferLineGetElements", function() M.get_elements() end, {})
```
akinsho, line 232: I don't think a command is useful here since this functionality will need to be used in a function, this command essentially does nothing but also can't be used to do anything really

akinsho (commented):
This looks good @andy-bell101 I definitely prefer this approach, just made a small comment about using the map function and removing the command. Another thing would be ensuring that group indicators don't get included in the elements

lua/bufferline/commands.lua
andy-bell101: Done :)

README.md
```diff
@@ -291,6 +292,37 @@ This order can be persisted between sessions (enabled by default).
 
 see `:help bufferline-custom-areas`
 
+#### Working with Elements
```
akinsho, line 295: @andy-bell101 was just about to merge this when I noticed this explanation was in the README as well. I've actually moved everything other than just cursory videos and eye candy stuff into the help docs. The README now is basically just an annotated gallery so this bit should just be in the help docs alone

akinsho (changes requested):
@andy-bell101 sorry didn't notice earlier that you'd added the explanation to the README but I actually don't put things like this in the README anymore just the help docs. I want the help docs to be where people go to figure out how the plugin works and the README is essentially an ad with shiny pictures 😆

andy-bell101:
No worries should be gone now

akinsho (approved):
LGTM 👍🏿 thanks for your work on this

andy-bell101:
No problem man, thanks for the plugins! Better sure I use a few of yours

PR #587, opened 2022-10-13
fix(commands): trigger redraw after delete in direction

### LINKED ISSUE #586: [Bug]: Bufferline does not refresh if BufferLineCloseLeft and BufferLineCloseRight are invoked using a keybinding

### What happened?

When I directly use `BufferLineCloseLeft` or `BufferLineCloseRight` commands it closes the buffers to the left or right respectively and the bufferline is refreshed/redrawn to reflect those changes however when I invoke those commands using a keybinding, in my case `<leader>b[` and `<leader>b]` respectively then it does the job of closing the buffers however the bufferline is not redrawn to show the changes unless a new action or keybinding is invoked(just pressing the leader key does the job) however if I just keep on navigating up/down inside the buffer then the bufferline is not redrawn to show the changes.

Bufferline should redraw as soon as the keybinding is invoked since it works without the keybinding.

### Config

local status_ok, bufferline = pcall(require, "bufferline")
if not status_ok then return end
bufferline.setup(tnvim.user_plugin_opts("plugins.bufferline", {
  options = {
    offsets = {
      { filetype = "NvimTree", text = "File Tree", highlight = "PanelHeading", padding = 1 },
      { filetype = "neo-tree", text = "File Tree", highlight = "PanelHeading", padding = 1 },
      { filetype = "Outline", text = "", padding = 1 },
    },
    show_close_icon = true,
    max_name_length = 14,
    max_prefix_length = 13,
    tab_size = 20,
    separator_style = "thin",
    show_duplicate_prefix = false,
  },
}))

### Additional Information

NVIM v0.8.0
OS: Arch Linux

### commit

_No response_

nat212:
Hi! I made a PR to fix the bufferline not redrawing after calling `close_in_direction` from a keybinding :)

This will fix #586

### DIFF
--- a/lua/bufferline/commands.lua
+++ b/lua/bufferline/commands.lua
@@ -206,6 +206,7 @@ function M.close_in_direction(direction, force)
       delete_element(item.id, force)
     end
   end
+  ui.refresh()
 end
 
 --- sorts all elements


akinsho:
@nat212 thanks for picking this up. I'd been meaning to get to this at some point

akinsho:
Thanks again 👍🏿 I've merged this into the `dev` branch which is where I put changes before doing a proper tagged release into main. I plan on doing a v3.0.0 release in a few days so should hit `main` soon.

PR #588, opened 2022-10-16
feat(tabs): allow moving/reordering tabs

toh995:
Before this PR, when in tab mode, the commands `BufferLineMoveNext` or `BufferLineMovePrev` did not visually change any tab positionings in the UI. Now it will update in the UI.

### Before:

https://user-images.githubusercontent.com/52012721/196027505-feeac011-3cfe-4773-a031-dc9ff742fcf7.mov

### After:

https://user-images.githubusercontent.com/52012721/196027539-84d08805-0af8-4f7c-b04a-48e538780389.mov

### DIFF
--- a/lua/bufferline/tabpages.lua
+++ b/lua/bufferline/tabpages.lua
@@ -75,27 +75,6 @@ local function get_valid_tabs()
   )
 end
 
---- sorts tab_nums in place, but doesn't add/remove any values
---- @param tab_nums number[]
---- @param sorted number[]
---- @return number[]
-local function get_updated_tabs(tab_nums, sorted)
-  if not sorted then return tab_nums end
-  local nums = { unpack(tab_nums) }
-  local reverse_lookup_sorted = utils.tbl_reverse_lookup(sorted)
-
-  --- a comparator that sorts buffers by their position in sorted
-  local sort_by_sorted = function(buf_id_1, buf_id_2)
-    local buf_1_rank = reverse_lookup_sorted[buf_id_1]
-    local buf_2_rank = reverse_lookup_sorted[buf_id_2]
-    if not buf_1_rank then return false end
-    if not buf_2_rank then return true end
-    return buf_1_rank < buf_2_rank
-  end
-  table.sort(nums, sort_by_sorted)
-  return nums
-end
-
 ---Filter the buffers to show based on the user callback passed in
 ---@param buf_nums integer[]
 ---@param callback fun(buf: integer, bufs: integer[]): boolean
@@ -138,7 +117,7 @@ end
 function M.get_components(state)
   local options = config.options
   local tab_nums = get_valid_tabs()
-  tab_nums = get_updated_tabs(tab_nums, state.custom_sort)
+  tab_nums = utils.apply_sort(tab_nums, state.custom_sort)
 
   local Tabpage = models.Tabpage
   ---@type NvimTab[]
--- a/lua/bufferline/tabpages.lua
+++ b/lua/bufferline/tabpages.lua
@@ -75,6 +75,27 @@ local function get_valid_tabs()
   )
 end
 
+--- sorts tab_nums in place, but doesn't add/remove any values
+--- @param tab_nums number[]
+--- @param sorted number[]
+--- @return number[]
+local function get_updated_tabs(tab_nums, sorted)
+  if not sorted then return tab_nums end
+  local nums = { unpack(tab_nums) }
+  local reverse_lookup_sorted = utils.tbl_reverse_lookup(sorted)
+
+  --- a comparator that sorts buffers by their position in sorted
+  local sort_by_sorted = function(buf_id_1, buf_id_2)
+    local buf_1_rank = reverse_lookup_sorted[buf_id_1]
+    local buf_2_rank = reverse_lookup_sorted[buf_id_2]
+    if not buf_1_rank then return false end
+    if not buf_2_rank then return true end
+    return buf_1_rank < buf_2_rank
+  end
+  table.sort(nums, sort_by_sorted)
+  return nums
+end
+
 ---Filter the buffers to show based on the user callback passed in
 ---@param buf_nums integer[]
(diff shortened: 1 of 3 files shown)


lua/bufferline/tabpages.lua
```diff
@@ -75,6 +75,27 @@ local function get_valid_tabs()
...
+  local sort_by_sorted = function(buf_id_1, buf_id_2)
+    local buf_1_rank = reverse_lookup_sorted[buf_id_1]
+    local buf_2_rank = reverse_lookup_sorted[buf_id_2]
+    if not buf_1_rank then return false end
+    if not buf_2_rank then return true end
+    return buf_1_rank < buf_2_rank
+  end
+  table.sort(nums, sort_by_sorted)
+  return nums
+end
```
toh995, line 97: Note that I just copied this function [from the buffers file](https://github.com/akinsho/bufferline.nvim/blob/f4dbf0b/lua/bufferline/buffers.lua#L23-L42).

I'm open to making this more DRY, i.e. define this function once and using it in both places... Please advise where the common function should be defined 😅

akinsho: @toh995 thanks for the PR. This generally looks good. It would definitely be cool to have this as a common utility. You can put this in the main utils file and import it in both places if you can make it more general

toh995: Thanks for the feedback @akinsho!

Per your recommendation, I added a commit extracting the sort function into the main utils file.

lua/bufferline/utils/init.lua
```diff
@@ -147,6 +147,28 @@ function M.get_tab_count() return #fn.gettabinfo() end
...
+  local nums = { unpack(to_sort) }
+  local reverse_lookup_sorted = M.tbl_reverse_lookup(sorted)
+
+  --- a comparator that sorts buffers by their position in sorted
+  local sort_by_sorted = function(buf_id_1, buf_id_2)
+    local buf_1_rank = reverse_lookup_sorted[buf_id_1]
+    local buf_2_rank = reverse_lookup_sorted[buf_id_2]
+    if not buf_1_rank then return false end
+    if not buf_2_rank then return true end
+    return buf_1_rank < buf_2_rank
```
akinsho, line 166: Thanks for splitting it out 🙌🏿 seems these variable names don't make sense any more, though. Think it would be better to go with `first_item`, `first_item_rank` etc. or anything more general just not directly referencing buffers

toh995: Good catch @akinsho! I updated.

akinsho (approved):
LGTM 👍🏿 thanks

toh995:
Hey @akinsho thanks for merging!

Is this supposed to be included on the 3.0.0 release notes? https://github.com/akinsho/bufferline.nvim/releases/tag/v3.0.0

akinsho:
@toh995 yeah they should have been. It didn't autogenerate the release notes this time so I did it by hand and forgot some things. I'll update it later 👍🏾

rwblokzijl:
@toh995 I think this broke `BufferLineSortByTabs`.

rwblokzijl:
I fixed it by adding the following to my config:
```
    sort_by = function(buffer_a, buffer_b)
        return buffer_a.ordinal < buffer_b.ordinal
    end,

```

toh995:
Hey @rwblokzijl I apologize for breaking some existing functionality!

I think it'll be hard to predict whether I'll be able to carve out some free time to investigate over the next few weeks.

@akinsho As a quick fix, maybe we could revert this PR for now? What do you think?

akinsho:
I'm fine to revert it, if either of you have a long term preferable solution feel free to raise a PR for that

toh995:
Hey @rwblokzijl just a heads up, the revert PR is merged! https://github.com/akinsho/bufferline.nvim/pull/590

Sorry for the inconvenience.

rwblokzijl:
@toh995 thanks for that! Btw, i think [this](https://github.com/akinsho/bufferline.nvim/pull/588#issuecomment-1282208792) might solve your issue too. Maybe that should even be default behaviour if `mode = "tabs"`.

@akinsho what was the original purpose of `BufferLineSortByTabs` and could it be depricated if [this](https://github.com/akinsho/bufferline.nvim/pull/588#issuecomment-1282208792) became default behaviour?

akinsho:
@rwblokzijl so sort by tabs was added by a user a while ago as a way of grouping buffers inside the same tab close to each other (IIRC). It isn't really intended to be used when using tabs mode since it won't make sense in that case.

Not sure that it needs to be deprecated, since it doesn't really relate to tab mode functionality at all. Also keep in mind that if a user sets the mode to tabs then we actually automatically change the sorting function to sort by tab id unless the user has specified something else

rwblokzijl:
@akinsho I had a look at the code and it seems to sort by tab id (neovim internal handle) instead of tab number (vim user sorted tabpages). Bufferlines `sort_by_tabs` already properly sorts this. I made the following PR to apply it by default: https://github.com/akinsho/bufferline.nvim/pull/592

adoyle-h:
Hi all, the `BufferLineMovePrev`/`BufferLineMoveNext` not work for me when `mode = 'tabs'`. Is there any problem in my config?

My bufferline.nvim git commit is 028a8793 (tag: v3.1.0).

nvim version:

```
NVIM v0.8.0
Build type: Release
LuaJIT 2.1.0-beta3
```

I have disabled all plugins and only require bufferline:

```lua
require('bufferline').setup({
  mode = 'tabs',
})
```

akinsho:
@adoyle-h this change was reverted because it caused a bug

rwblokzijl:
@adoyle-h the builtin `gt` and `gT` should do the things you want to do.

toh995:
Hey @adoyle-h `:tabmove+` and `:tabmove-` seem to be working for me, using bufferline tab mode.

adoyle-h:
@toh995 It works. Thank you.

Issue #549, opened 2022-09-14
[Feature Request]: Show icon based on filetype instead of file extension


4e554c4c:
### What?

Sometimes two files which have the same extension have different filetypes.
I see this very often because I often write [hack](https://hacklang.org/) which can have the `.php` filetype even though it is an entirely different programming language.

This also occurs in vim when there is no filename attached to a buffer, but it still has a filetype.

### Why?

I often find myself editing buffers (either through a buffer without an attached file or a file with a different filetype than is ordinary) which have a different filetype than that associated with its filename's extension. I would like to be able to still have the correct icon shown in these situations without overriding the file extension globally.

akinsho:
@4e554c4c please try the latest commit of the `dev` branch, this should now be fixed. I'll release this when next I get a chance

PR #598, opened 2022-10-28
fix(utils): get icon by filetype when get icon by filename not exist

teddy2104:
I think get icon by filetype should be fallback of get icon by filename/extension.
Before:
before
After:
after

### DIFF
--- a/lua/bufferline/utils/init.lua
+++ b/lua/bufferline/utils/init.lua
@@ -186,15 +186,12 @@ function M.get_icon(opts)
 
   local use_default = config.options.show_buffer_default_icon
 
-  local icon, hl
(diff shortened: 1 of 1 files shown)


akinsho:
@teddy2104 thanks for the PR, although I'm pretty sure this will break the use case of the user who prompted me to add in this change in the first place. The path will almost always exist so it will almost never fall through to the `filetype` check. This issue https://github.com/akinsho/bufferline.nvim/issues/549 lead to me adding this change and the problem was that files can have the same extension but different filetypes. Unfortunately, this PR would break those cases (again)

teddy2104:
Yes, it's all right. Thank you for this great plugin.

PR #603, opened 2022-11-01
Add a new separator to show hidden built-in groups

xwjdsh:
image

This is an example of my buffers, when I hid the built-in `ungrouped` and `pinned` groups, they disappeared. It's a bit confusing, and after a while, people may forget the hidden groups.

And after the changes,

image

### DIFF
--- a/lua/bufferline/groups.lua
+++ b/lua/bufferline/groups.lua
@@ -114,6 +114,16 @@ end
 ---@type GroupSeparator
 function separator.none() return { sep_start = {}, sep_end = {} } end
 
+---@param group Group,
+---@param hls  table<string, table<string, string>>
+---@param count string
+---@return Separators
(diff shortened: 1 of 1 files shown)


akinsho:
@xwjdsh builtin groups like pinned and ungroup are not supposed to have indicators, you are not supposed to hide these groups. That they are actually groups is an implementation detail that a user should not be depending on, as their implementation could possibly change. So I don't expect to have to show them if they are hidden (since they should not be being hidden)

xwjdsh:
image

When I press the tab, it shows in the pop-up window (then I supposed that's OK to toggle built-in groups), and I think that's useful to toggle/close built-in groups in some cases, such as I can pin some buffers and collapse other ungrouped, then I have a clear buffer line.

PR #606, opened 2022-11-04
make example config into valid lua

stevenctl:
This should allow copying and pasting directly from the docs.

The main problem I see with this PR is it might not be setting the defaults (although I personally like what it ended up being). I was just making it set the first example in the `|` delimited list of options. 

When configuring a plugin, I like to paste the entire default config into my own to use as a template, which is why I sent the PR.

### DIFF
--- a/doc/bufferline.txt
+++ b/doc/bufferline.txt
@@ -66,7 +66,7 @@ The available configuration are:
     require('bufferline').setup {
         options = {
             mode = "buffers", -- set to "tabs" to only show tabpages instead
-            numbers = "none" | "ordinal" | "buffer_id" | "both" | function({ ordinal, id, lower, raise }): string,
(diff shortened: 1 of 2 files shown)


akinsho (commented):
@stevenctl thanks for the PR. To be honest, I'm really against people copy pasting large blocks of code that they haven't read into their configs. This I think is the reason for a bunch of issues because rather than search a topic in the help docs then read and understand it. Instead, they copy and paste everything and then don't know what stuff does.

In this case, a user should definitely not copy all of this config ever because no single user needs to override every single field. If they just use `bufferline.setup()` they already get all the defaults. If they want to change something, then they *should* read the help docs.

I think if this gets merged it will be quickly outdated because a new contributor who adds a new option might forget a comma or something, but users will think that it is valid to copy and paste this, and then I'll get a bug report about invalid syntax

stevenctl:
Fair enough.

PR #616, opened 2022-11-14
fix(hover): check showtabline not laststatus

vimpostor:
There is not really a point in gatekeeping this feature from `laststatus=0`.
The hover feature works completely fine with `laststatus=0` and this creates only problems with other plugins, see: https://github.com/vimpostor/vim-tpipeline/issues/39

### DIFF
--- a/lua/bufferline/hover.lua
+++ b/lua/bufferline/hover.lua
@@ -10,7 +10,7 @@ local previous_pos = nil
 local M = {}
 
 local function on_hover(current)
-  if vim.o.laststatus == 0 then return end
+  if vim.o.showtabline == 0 then return end
   if current.screenrow == 1 then
     api.nvim_exec_autocmds("User", {
       pattern = "BufferLineHoverOver",


lua/bufferline/hover.lua
```diff
@@ -10,7 +10,6 @@ local previous_pos = nil
 local M = {}
 
 local function on_hover(current)
-  if vim.o.laststatus == 0 then return end
```
akinsho, line 13: That was supposed to be `showtabline` not `laststatus`

akinsho (changes requested):
@vimpostor would be great if you could tweak this PR to add a check for whether the tabline is showing or not, which was my original intention there.

lua/bufferline/hover.lua
vimpostor: Done

akinsho:
Thanks for the contribution 👍🏿

Issue #619, opened 2022-11-16
[Bug]:  Padded slanted separators breaks the separators in SimpleTerminal (st)


papitz:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

In st with JetBrainsMono as the font the padded_slant option breaks the separators. Normal slant also causes problems but thats what the padding is for I guess. 
The separators look like this:
![padded_slant_not_fixed](https://user-images.githubusercontent.com/65065967/202197533-6b62422f-2820-4984-a68b-fc8a12b9443e.png)
This can be fixed by adding the padding on the slanted side of the separator instead of on the right both times.
It makes it look like this:
![padded_slant_fixed](https://user-images.githubusercontent.com/65065967/202197741-f7ecfc0d-c071-46fd-9192-80415a0acb26.png)

### What did you expect to happen?

Looking like this:

![padded_slant_fixed](https://user-images.githubusercontent.com/65065967/202197798-e0ba3940-0d1a-4350-ab1a-5d8ca84f9720.png)

### Config
```
require("bufferline").setup {
    options = {
        separator_style = { "padded_slant" },
    },
}
```

### Additional Information

I'll submit a pull request, but it might be possible that this would affect how the separators look in other terminals or with other fonts.

PR #620, opened 2022-11-16
fix: fixed separator padding

papitz:
Regarding issue #619


akinsho:
Unfortunately, this will almost certainly break for a bunch of other people, and I don't have the bandwidth to try and maintain support for every terminal emulator. If you can test and validate with screenshots here that this does not break on Kitty, Alacritty, Gnome Terminal etc. then I'll consider merging, but frankly I doubt this won't break for somebody. As I mentioned in the readme, this style won't work for everyone, and I don't want to be in the business of trying to accommodate universal support.

papitz:
Alright makes sense. I'll see if I can test some more terminals but I think you're right about it breaking in other terminals. For now I'll just use one of the other styles or fix it locally since I really enjoy the plugin! Thanks a lot

PR #635, opened 2022-12-15
add tabpage renaming

### LINKED ISSUE #634: [Feature Request]: Tab naming
### What?

It would be nice to have named tabs instead of the current ` 1 | 2 | 3 `

i.e. ` project 1 | project 2 | nvim-conf `

Something like a `BufferlineTabRename` or similar?

### Why?

If you're a tab _and_ buffer user, allows for less thinking about which tab has what during a long session or coming back to nvim after a period of being away from the desk.

tabline.nvim has this already, but I very much prefer the existing feature set of bufferline.nvim

Haven't dug in to the code yet, so I'm not sure how much this would take. But I can maybe submit a PR if it's not too bad.

distek:
Fixes #634 

The name can be set like so:
```lua
-- Set's current tab to "something"
require('bufferline.tabpages').rename_tab(0, "something")
-- Set's tab 1 to "asdf"
require('bufferline.tabpages').rename_tab(1, "asdf")
```
or
```vim
" both of these are equivalent 
:BufferLineRenameTab something
" or
:BufferLineRenameTab 0 something
```
All remaining arguments to `BufferLineRenameTab` are taken as the name:
```vim
" This will rename tab 1 to "this is project 1"
:BufferLineRenameTab 1 this is project 1
```

Example result:
bufferline-tabline-rename

### DIFF
--- a/lua/bufferline/tabpages.lua
+++ b/lua/bufferline/tabpages.lua
@@ -41,9 +41,10 @@ function M.rename_tab(tabnr, name)
   if tabnr == 0 then
     tabnr = vim.fn.tabpagenr()
   end
-
+  if name == "" then
+    name = ""..tabnr..""
+  end
   vim.api.nvim_tabpage_set_var(tabnr, "name", name)
-
   M.get()
 end
 
--- a/lua/bufferline/tabpages.lua
+++ b/lua/bufferline/tabpages.lua
@@ -30,11 +30,21 @@ local function render(tabpage, is_active, style, highlights)
   local separator_hl = is_active and h.tab_separator_selected.hl_group or h.tab_separator.hl_group
   local chars = constants.sep_chars[style] or constants.sep_chars.thin
   local separator_component = chars[2]
-  local name = padding .. padding .. tabpage.tabnr .. padding
+  local name = padding .. padding .. (tabpage.variables.name or tabpage.tabnr) .. padding
   return {
     { highlight = hl, text = name, attr = { prefix = tab_click_component(tabpage.tabnr) } },
     { highlight = separator_hl, text = separator_component },
   }
+end
+
+function M.rename_tab(tabnr, name)
+  if tabnr == 0 then
+    tabnr = vim.fn.tabpagenr()
+  end
+
+  vim.api.nvim_tabpage_set_var(tabnr, "name", name)
+
+  M.get()
 end
 
 function M.get()
--- a/doc/bufferline.txt
+++ b/doc/bufferline.txt
@@ -191,7 +191,7 @@ You can change the appearance of the bufferline separators by setting the
 
 
 ==============================================================================
-TABPAGES                                                          *bufferline-tabs*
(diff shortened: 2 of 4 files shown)


lua/bufferline/tabpages.lua
```diff
@@ -30,13 +30,24 @@ local function render(tabpage, is_active, style, highlights)
   local separator_hl = is_active and h.tab_separator_selected.hl_group or h.tab_separator.hl_group
   local chars = constants.sep_chars[style] or constants.sep_chars.thin
   local separator_component = chars[2]
-  local name = padding .. padding .. tabpage.tabnr .. padding
+  local name = padding .. padding .. (tabpage.variables.name or tabpage.tabnr) .. padding
```
AlexSWall, line 33: I would have thought `tabpage.tabnr` would be just an integer, which suggests to me (without good knowledge of the codebase) that this line isn't actually replacing the standard contents of the tab, but just the number? Does this actually work? This change seems like the crux of all I'm uncertain of in this PR.

lua/bufferline/tabpages.lua
```diff
@@ -30,13 +30,24 @@ local function render(tabpage, is_active, style, highlights)
...
     { highlight = separator_hl, text = separator_component },
   }
 end
 
+function M.rename_tab(tabnr, name)
+  if tabnr == 0 then
+    tabnr = vim.fn.tabpagenr()
+  end
+  if name == "" then
+    name = ""..tabnr..""
```
AlexSWall, line 45: Seems you handle defaulting to `tabnr` above, so instead we can set the variable to `nil`?
```suggestion
    name = nil
```

AlexSWall (commented):
I'm keen for this functionality so taking a quick look as an interested third party.

lua/bufferline/tabpages.lua
AlexSWall: So this wouldn't work for me, but instead you can pass in the following `name_formatter` to `opts`:
```lua
name_formatter = function(inputs)
    local tab_info = vim.fn.gettabinfo(inputs.tabnr)[1]
    return tab_info.variables.name or inputs.name
end
```
This then all works for me with my setup (albeit doesn't immediately refresh properly; I need to change tab first).

distek:
Hey @AlexSWall

I'm actually going to close this for a couple of reasons: akinsho didn't seem to have a whole lot of interest in it, and I've found myself moving away from neovim in the terminal anyway. If you'd like to take this over, please do!

muhmud:
I wouldn't mind taking this over, if you were interested in the change @akinsho 

I think with the `scope.nvim` integration, this would be a nice addition.

akinsho:
@muhmud I don't really mind at all and am not particularly engaged in the tab page stuff (short of reviewing +/- merging) for the most part I don't use them but many dedicated tab users have contributed bits of functionality over time, From my end if the code is well implemented and works without breaking any tests (and maybe even adds some) then all good by me.

PR #636, opened 2022-12-18
Add ability to cycle through buffers which are not currently open in …

roobert:
…a window

Untitled

### DIFF
--- a/doc/bufferline.txt
+++ b/doc/bufferline.txt
@@ -693,6 +693,11 @@ A few of this plugins commands can be mapped for ease of use. >
     nnoremap <silent>[b :BufferLineCycleNext<CR>
     nnoremap <silent>b] :BufferLineCyclePrev<CR>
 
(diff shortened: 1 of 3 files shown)


akinsho:
@roobert thanks for the PR 👍🏿. Tbh I'm not sure why this functionality is necessary. It seems like it might relate to a specific workflow you have? I'm not sure why we'd need a specific set of commands for avoiding open buffers? I'm wondering if this can't be accomplished with the exposed functionality from this plugin already without having to be built in as well

roobert:
Fair points. I think one of the  barriers to entry to Neovim is vims buffers/tabs model which your plugin goes some way to improving the experience for. However, when coming from conventional apps having two views of the same file can be unexpected and potentially undesired. I thought I'd open this PR to give people another option that may feel more natural but also happy to implement this locally if you'd prefer not to merge. 👍

roobert:
I've implemented this as a separate extension so closing this PR 👍  - https://github.com/roobert/bufferline-cycle-windowless.nvim

PR #642, opened 2022-12-23
Update bufferline.txt

lodisy:
Add missing commas at L96 & L128


akinsho:
@lodisy I've actually received a variant of this PR before but explained to the author that the code you are adding commas two is not valid lua *anyway* it is there for the purposes of explanation *not for copy and pasting* I personally don't maintain or add any code blocks in any of my plugins that are designed for copy and pasting since I strongly believe people should read the code and write what they specifically need themselves as copy pasting just creates more work for me in the long run because people didn't read the docs anyway

PR #643, opened 2022-12-23
fix(highlights): use correct background

casonadams:


casonadams:
I am not quite sure what broke this test.  Would love to help debug if I can get some more details.

akinsho:
@casonadams can you run the tests locally using `plenary` and see if there are actual failures locally. I think what you are seeing in CI here is an unrelated error.

casonadams:
@akinsho running tests locally everything is passing.  How do I move forward?

akinsho:
@casonadams that's good, nothing on your end I'll test the changes locally when I have time and merge them off they're ok

PR #646, opened 2022-12-24
refactor(tabs): only show close icon if multiple tabse

dwijesinghe:
The tab close icon can look a little odd (and won't do anything) if there's only one tab in use. See [this reddit post ](https://www.reddit.com/r/neovim/comments/v9211b/can_we_remove_the_x_on_the_right_side_of_the/)

Since I would guess many people who use this plugin would be less inclined to use standard vim tabs - I thought it would make sense and improve aesthetics to remove it if there's only one tab in use (it will reappear if the user opens another tab). For those who want to always see the tab close icon, the `always_show_tab_close_icon` option can be set to `true` in setup (default is `false`). 

Please let me know your thought and btw I love this plugin :)

### DIFF
--- a/doc/bufferline.txt
+++ b/doc/bufferline.txt
@@ -537,7 +537,7 @@ for left or right aligned sidebar windows such as `NvimTree`, `NERDTree` or
         {
             filetype = "NvimTree",
             text = "File Explorer",
-            highlight = "Directory"
+            highlight = "Directory",
             separator = true -- use a "true" to enable the default, or set your own character
         }
     }
--- a/doc/bufferline.txt
+++ b/doc/bufferline.txt
@@ -150,6 +150,7 @@ The available configuration are:
             sort_by = 'insert_after_current' |'insert_at_end' | 'id' | 'extension' | 'relative_directory' | 'directory' | 'tabs' | function(buffer_a, buffer_b)
                 -- add custom logic
                 return buffer_a.modified > buffer_b.modified
+            always_show_tab_close_icon = true | false, -- if false the tab close icon will only appear if there is more than one tab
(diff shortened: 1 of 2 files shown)


akinsho:
@dwijesinghe this is a good idea, thanks for the PR. I had been considering something similar, but hadn't gotten round to implementing it. The only suggestion I have would be to make this behaviour the default also I'd go as far as to remove the option tbh since frankly it's useless to have a button that does nothing and I get issues from users who are confused that the button "does not work" when there is only one tab open. I'd say just change this to be the only behaviour, i.e. only show the button if there is a tab.

doc/bufferline.txt
```diff
@@ -150,6 +150,7 @@ The available configuration are:
             sort_by = 'insert_after_current' |'insert_at_end' | 'id' | 'extension' | 'relative_directory' | 'directory' | 'tabs' | function(buffer_a, buffer_b)
                 -- add custom logic
                 return buffer_a.modified > buffer_b.modified
+            always_show_tab_close_icon = true | false, -- if false the tab close icon will only appear if there is more than one tab
```
akinsho, line 153: I'd say don't make this optional at all, I made a mistake by offering too many things as options. I'd say to just be decisive and make this the only behaviour since the alternative is useless

akinsho (changes requested):
Made a couple of comments re. the implementation but definitely like this idea

dwijesinghe:
@akinsho Thanks! Just made those changes, its def a lot more succinct now. Let me know if there's anything else you want modified etc.

akinsho (approved):
LGTM

PR #653, opened 2023-01-03
fix(get_icon): always use `get_icon` API of devicons

utilyre:
This API has better icon matching. As an example, `package.json` would have a generic json icon before this commit, but now it gets the specific PackageJson icon and highlight.

Before:

before

After:

after

### DIFF
--- a/lua/bufferline/utils/init.lua
+++ b/lua/bufferline/utils/init.lua
@@ -184,6 +184,9 @@ function M.get_icon(opts)
   end
   if type == "terminal" then return webdev_icons.get_icon(type) end
 
+  -- NOTE(#549): hack filetype might use the same extension as php, so prioritize filetype based icon look up in that case
+  if opts.filetype == "hack" then return webdev_icons.get_icon_by_filetype(opts.filetype) end
+
   local use_default = config.options.show_buffer_default_icon
   local icon, hl = webdev_icons.get_icon(fn.fnamemodify(opts.path, ":t"), opts.extension, {
     -- Don't use a default here so that we fall through to the next case if no icon is found
--- a/lua/bufferline/utils/init.lua
+++ b/lua/bufferline/utils/init.lua
@@ -185,17 +185,9 @@ function M.get_icon(opts)
   if type == "terminal" then return webdev_icons.get_icon(type) end
 
   local use_default = config.options.show_buffer_default_icon
-
-  local icon, hl
-  if M.is_truthy(opts.filetype) then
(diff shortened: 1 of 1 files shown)


akinsho:
@utilyre this specific issue has come up before at least twice now. I only moved to using the filetype because a user using php kept getting the same icon even though hack and php have the same extension but different filetypes or something like that. Tbh I can't remember the specifics and I personally am not bothered by the issue he was, but this PR would essentially regress the behaviour for that user and cause issues all over again 😮‍💨.

Ultimately if there is no way to not break other people's use cases than I'd prefer this mechanism but a better no breaking solution would be preferred.

utilyre:
I see. I looked up `hack`, `hk`, and `php` in [web-devicons](https://github.com/nvim-tree/nvim-web-devicons) and there seems to be nothing related to a `hack` filetype.

So assuming that #549 was using the `override` option in `devicons.setup()`, his icon would be inside the `icons` table of [web-devicons](https://github.com/nvim-tree/nvim-web-devicons) and not the `filetypes` table (which is being used by `*_by_filetype` helpers).

Nonetheless, we can prevent breaking changes by a simple if statement for exceptionally using `get_icon_by_filetype` for a `php` extension, I suppose.

utilyre:
The last patch I applied covers the `hack` ft case.

lua/bufferline/utils/init.lua
```diff
@@ -184,17 +184,16 @@ function M.get_icon(opts)
   end
   if type == "terminal" then return webdev_icons.get_icon(type) end
 
-  local use_default = config.options.show_buffer_default_icon
+  -- NOTE(#549): hack filetype might use the same extension as php, so prioritize filetype based icon look up in that case
+  if opts.filetype == "hack" then return webdev_icons.get_icon_by_filetype(opts.filetype) end
```
akinsho, line 188: @utilyre I would rather not hard code special cases like this into the code base at all. I'd rather break this for all special cases or provide users a way to handle this through some sort of official flexible mechanism

akinsho (changes requested):

camiloaromero23:
@utilyre @akinsho Just submitted a PR that addresses this issue from other point of view. I believe this is a better alternative that will not affect none of the users of the plugin and also will give the possibility to use the `get_icon` devicons API. [Get Icon Alternative PR](https://github.com/akinsho/bufferline.nvim/pull/669)

utilyre:
Closed this in favor of #669

PR #659, opened 2023-01-16
Feature: add slope separator 

### LINKED ISSUE #658: [Feature Request]: Add option for custom 'slope' seperator
### What?

Hoping to have the ability to style the separators similar to this  
image

image

### Why?

Some people like the sloped look and it seems the simplest option would be to update the sep_options to have a new option 'slope'

Taaqif:
Should close #658

### DIFF
--- a/doc/bufferline.txt
+++ b/doc/bufferline.txt
@@ -184,8 +184,6 @@ You can change the appearance of the bufferline separators by setting the
 * `padded_slant` - Same as `slant` but with extra padding which some terminals require.
   If `slant` does not render correctly for you try padded this instead.
 * `slope` - Use slanted/triangular separators but slopped to the right
-* `padded_slope` - Same as `slope` but with extra padding which some terminals require.
-  If `slope` does not render correctly for you try padded this instead.
 * `thick` - Increase the thickness of the separator characters
 * `thin` - (default) Use thin separator characters
 * finally you can pass in a custom list containing 2 characters which will be
--- a/doc/bufferline.txt
+++ b/doc/bufferline.txt
@@ -139,7 +139,7 @@ The available configuration are:
             persist_buffer_sort = true, -- whether or not custom sorted buffers should persist
             -- can also be a table containing 2 custom separators
             -- [focused and unfocused]. eg: { '|', '|' }
-            separator_style = "slant" | "thick" | "thin" | { 'any', 'any' },
+            separator_style = "slant" | "slope" | "thick" | "thin" | { 'any', 'any' },
             enforce_regular_tabs = false | true,
             always_show_bufferline = true | false,
             hover = {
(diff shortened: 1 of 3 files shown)


doc/bufferline.txt
```diff
@@ -183,6 +183,9 @@ You can change the appearance of the bufferline separators by setting the
 * `slant` - Use slanted/triangular separators
 * `padded_slant` - Same as `slant` but with extra padding which some terminals require.
   If `slant` does not render correctly for you try padded this instead.
+* `slope` - Use slanted/triangular separators but slopped to the right
+* `padded_slope` - Same as `slope` but with extra padding which some terminals require.
```
akinsho, line 187: @Taaqif do you know if this is actually needed or have you added it because you saw another style which used it. I'd rather not add this style as the original was a hack to fix the rendering of the slant in some terminals

Taaqif: Hi @akinsho 

Yeah I dont think this is really needed. I only added it as I copied the slant config. 
Happy to remove it as my tests showed the slope (without padding) works just fine on my terminal emulator 

Your call 👍 , I'll update the PR if you prefer

akinsho: @Taaqif yes please it might turn out to be necessary but I'd rather start with it not being there and wait to see if it ever comes up

Taaqif: All done!

Thanks for the guidance 😄

lua/bufferline/constants.lua
```diff
@@ -11,8 +11,6 @@ M.sep_names = {
   thick = "thick",
   slant = "slant",
   slope = "slope",
-  padded_slant = "padded_slant",
```
akinsho, line 13: @Taaqif you've removed more than just the one you've added, you've actually removed the `padded_slant` which is definitely needed

lua/bufferline/constants.lua
```diff
@@ -21,8 +19,6 @@ M.sep_chars = {
   [M.sep_names.thick] = { "▌", "▐" },
   [M.sep_names.slant] = { "", "" },
   [M.sep_names.slope] = { "", "" },
-  [M.sep_names.padded_slant] = { "" .. M.padding, "" .. M.padding },
```
akinsho, line 21: Please don't remove this option as it's actually needed, just the padded variant you added 🙏🏿

akinsho (changes requested):
Woops looks like you've accidentally removed the existing padded variant

Taaqif:
> Woops looks like you've accidentally removed the existing padded variant

Oops @akinsho! 
Just re-added the padded_slant keys 🤕

akinsho (approved):
Tested this out locally and LGTM 👍🏿

PR #669, opened 2023-01-23
feat(icons): include load_icons_from_filetype option

camiloaromero23:
Included an option to avoid loading the devicons from the filetype in order to support user custom-defined devicons. Defaults to true for not breaking current configs


CleoMenezesJr (approved):

akinsho:
Just to mention I have seen this PR, and it generally looks fine to me except I really want to avoid the proliferation of options which has been an issue in this plugin. I want to solve this problem in a way where want to fetch the icon according to some preference or strategy can always be accounted for. I absolutely do not want a future request where a user wants to add in a `load_icons_from_buftype` or `load_icons_from_ext_with_custom_strategy`.

I'm thinking that a user should directly specify a function to use and if no function is provided the we just default to using the name.

e.g.

```lua
-- config 
{
  -- ... other opts, name is just an idea
  get_file_icon = require('webdevicons').get_by_method_i_prefer,
}

-- In the code

local icon_fn = config.options.get_file_icon or require('webdevicons').default_method)
local icon = icon_fn(params)

```

akinsho:
Closing this in favour of #686 I've ported over the commit from here but change the implementation to match my comment please have a go with that PR and let me know how it works 🙏🏿

PR #673, opened 2023-02-11
docs: update config type annotations

tobealive:
Adds missing types to config fields.

Though in case of the separator style id did just a quick scan through the code base, I can not 100% verify if the string dict is the right union type . But it seemed fine to me.


akinsho:
Thanks @tobealive 👍🏿

PR #675, opened 2023-02-14
feat: allow moving buffers to arbitrary positions

MisterFISHUP:
I added a `move_to` function and `BufferLineMoveTo{First|Last}` commands so one can easily move the current buffer to the `n`-th position in the bufferline.

It's a bit like `:tabmove` but for buffers.

`move_to` takes an 1-based buffer index. Negative integers are allowed but 0 is ignored.

---

P.S. This is an awesome project! Thank you to all the maintainers for putting in so much time and effort to make this possible. I'm really enjoying using it and I'm so grateful for all the work that went into it. Keep up the great work!"

### DIFF
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -35,6 +35,7 @@ local BUFFERLINE_GROUP = "BufferlineCmds"
 
 local M = {
   move = commands.move,
+  move_to = commands.move_to,
   exec = commands.exec,
   go_to = commands.go_to,
   cycle = commands.cycle,
@@ -233,6 +234,8 @@ local function setup_commands()
   cmd("BufferLineCloseLeft", function() M.close_in_direction("left") end, {})
   cmd("BufferLineMoveNext", function() M.move(1) end, {})
   cmd("BufferLineMovePrev", function() M.move(-1) end, {})
+  cmd("BufferLineMoveToFirst", function() M.move_to(1) end, {})
+  cmd("BufferLineMoveToLast", function() M.move_to(-1) end, {})
   cmd("BufferLineSortByExtension", function() M.sort_buffers_by("extension") end, {})
   cmd("BufferLineSortByDirectory", function() M.sort_buffers_by("directory") end, {})
   cmd(
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -234,8 +234,6 @@ local function setup_commands()
   cmd("BufferLineCloseLeft", function() M.close_in_direction("left") end, {})
   cmd("BufferLineMoveNext", function() M.move(1) end, {})
   cmd("BufferLineMovePrev", function() M.move(-1) end, {})
-  cmd("BufferLineMoveToFirst", function() M.move_to(1) end, {})
-  cmd("BufferLineMoveToLast", function() M.move_to(-1) end, {})
   cmd("BufferLineSortByExtension", function() M.sort_buffers_by("extension") end, {})
   cmd("BufferLineSortByDirectory", function() M.sort_buffers_by("directory") end, {})
   cmd(
--- a/lua/bufferline/commands.lua
+++ b/lua/bufferline/commands.lua
@@ -166,6 +166,23 @@ function M.move(direction)
   end
 end
 
+--- @param buffer_index number
+function M.move_to(buffer_index)
+  local index = M.get_current_element_index(state)
+  if not index then return utils.notify("Unable to find buffer to move, sorry", "warn") end
+  local next_index = buffer_index > 0 and buffer_index or #state.components + 1 + buffer_index
+  if next_index >= 1 and next_index <= #state.components then
+    local item = state.components[index]
+    local destination_buf = state.components[next_index]
+    state.components[next_index] = item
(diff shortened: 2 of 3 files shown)


lua/bufferline/commands.lua
```diff
@@ -166,6 +166,23 @@ function M.move(direction)
   end
 end
 
+--- @param buffer_index number
+function M.move_to(buffer_index)
```
akinsho, line 170: This function is almost entirely a duplicate of the above which is not very DRY. I would recommend implementing this by looking at how you can adapt the `move` funtion to take a specific argument. I would suggest just allowing the argument to be the buffer position and when move is for the existing commands just calculating the next index and the previous index

lua/bufferline.lua
```diff
@@ -233,6 +234,8 @@ local function setup_commands()
   cmd("BufferLineCloseLeft", function() M.close_in_direction("left") end, {})
   cmd("BufferLineMoveNext", function() M.move(1) end, {})
   cmd("BufferLineMovePrev", function() M.move(-1) end, {})
+  cmd("BufferLineMoveToFirst", function() M.move_to(1) end, {})
```
akinsho, line 237: I'm not sure that we need to proliferate the number of commands even further. I think if the move api is changed to allow just specifying a buffer index then a user can easily add this command themselves in their config and this can be easily documented in the README rather than adding more code to maintain here. Where some user might appear asking for move to second to the last or some other thing

akinsho (changes requested):
@MisterFISHUP thanks for the PR. I don't mind the idea of adapting the move function to allow specifying a position to move to, but the current solution is not DRY at all and I think it can be reworked. The  move function should be adapted to take an index and when used in the existing mappings the arguments should be calculated.

lua/bufferline/commands.lua
MisterFISHUP: Yes, I totally agree with you. You can review the updated PR. It's improved but now `get_current_element_index` is called twice in `move` and I'm not sure how to mitigate this problem, could you give me a clue? Thank you for your time.

akinsho: @MisterFISHUP you could pass the `current_element_index` as a nillable parameter and use it if passed and fetch if not e.g.
```lua
M.move_to(index + direction, index)

-- then in move to
function M.move_to(element_index, current_element)
 -- ... other stuff
 local index = current_element or M.get_current_element_index(state)
end
```

MisterFISHUP: @akinsho Thanks for the hint. `move_to` now takes `to_index` and `from_index`, which makes sense. I hope the implementation is cleaner. Tell me if the variable names are ok for you or you want something like source/target instead of from/to or else.

Speaking of docs, I'm not sure if I should show `move`, `move_to` (or add some examples of using them) in the readme or bufferline.txt, or maybe you want to update the doc later.

MisterFISHUP: @akinsho I added back the examples I wrote before, but with `move_to` command this time. I think this will be sufficient, people interested in the command can bind `move_to(2)` etc. themselves.

I didn't add examples where `move_to` takes the second argument `from_index` because I personally don't need it, but technically we can also do things like 'move the 4th buffer to the second position' now. What do you think? Do you think it's ok like this or it's better to add one or two examples to show that `move_to` can take an optional `from_index`?

Thank again for your time.

akinsho: @MisterFISHUP I think simple is better for now, we didn't really add that in for people to use though as you say it's possible but really there's no clean way to actually do that and it'll just invite issues/questions in the future which I'd rather avoid

lua/bufferline/commands.lua
```diff
@@ -149,11 +149,12 @@ function M.get_current_element_index(current_state, opts)
...
---- @param direction number
-function M.move(direction)
-  local index = M.get_current_element_index(state)
+--- @param to_index number
+--- @param from_index number?
+function M.move_to(to_index, from_index)
+  local index = from_index or M.get_current_element_index(state)
   if not index then return utils.notify("Unable to find buffer to move, sorry", "warn") end
-  local next_index = index + direction
+  local next_index = to_index > 0 and to_index or #state.components + 1 + to_index
```
akinsho, line 157: I'm not immediately sure what `#state.components + 1 + to_index` is doing here?

MisterFISHUP: `move_to` can accept negative indices (count from the right instead of the left): `move_to(-1)` moves the buffer to the last position, `-2` to the second-last, etc.

akinsho: @MisterFISHUP can you add a comment more or less saying that 🙏🏿 I'm bound to forget in a few months and not be sure why it's like that. After that this is good to go 🚀

MisterFISHUP: Yes, I hope it's clearer now :)

akinsho:
Thanks for all your work on this @MisterFISHUP much appreciated 👍🏿

MisterFISHUP:
Thank you for merging this PR, and for all the useful guides you have provided! 😊

PR #680, opened 2023-02-16
add go to buffer absolute

otavioschwanck:
To go direct to first with M-1 (or SPC 1), when i have a lot of buffers

### DIFF
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -242,6 +242,7 @@ local function setup_commands()
   )
   cmd("BufferLineSortByTabs", function() M.sort_buffers_by("tabs") end, {})
   cmd("BufferLineGoToBuffer", function(opts) M.go_to_buffer(opts.args) end, { nargs = 1 })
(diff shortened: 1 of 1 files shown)


akinsho:
Hi @otavioschwanck thanks for the PR but I don't think this is necessary. I actually actively deprecated the absolute buffer behaviour since a user can already create it easily in their own config and it is already documented.

In general I don't add new commands or options if they can be produced easily by the user since more code here is more work to maintain which makes it harder for me to keep this project going

otavioschwanck:
> Hi @otavioschwanck thanks for the PR but I don't think this is necessary. I actually actively deprecated the absolute buffer behaviour since a user can already create it easily in their own config and it is already documented.
> 
> In general I don't add new commands or options if they can be produced easily by the user since more code here is more work to maintain which makes it harder for me to keep this project going

I think the absolute is more important to the visible option, because of the pinned tabs.  If you need help to maintain, let me know.

akinsho:
Regardless this is easily done by creating the command in a users config and the argument re pinned tabs is quite specific to what might be useful in your workflow.

Issue #678, opened 2023-02-15
[Bug]: error when there are no tabs to close with `GroupClose`


otavioschwanck:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

```
Error executing Lua callback: .../nvim/lazy/bufferline.nvim/lua/bufferline/utils/init.lua:110: bad argument #1 to 'ipairs' (table expected, got nil)
stack traceback:
	[C]: in function 'ipairs'
	.../nvim/lazy/bufferline.nvim/lua/bufferline/utils/init.lua:110: in function 'for_each'
	...hare/nvim/lazy/bufferline.nvim/lua/bufferline/groups.lua:304: in function 'command'
	...local/share/nvim/lazy/bufferline.nvim/lua/bufferline.lua:135: in function 'group_action'
	...local/share/nvim/lazy/bufferline.nvim/lua/bufferline.lua:247: in function <...local/share/nvim/lazy/bufferline.nvim/lua/bufferline.lua:247>
Error executing Lua callback: .../nvim/lazy/bufferline.nvim/lua/bufferline/utils/init.lua:110: bad argument #1 to 'ipairs' (table expected, got nil)
stack traceback:
	[C]: in function 'ipairs'
	.../nvim/lazy/bufferline.nvim/lua/bufferline/utils/init.lua:110: in function 'for_each'
	...hare/nvim/lazy/bufferline.nvim/lua/bufferline/groups.lua:304: in function 'command'
	...local/share/nvim/lazy/bufferline.nvim/lua/bufferline.lua:135: in function 'group_action'
	...local/share/nvim/lazy/bufferline.nvim/lua/bufferline.lua:247: in function <...local/share/nvim/lazy/bufferline.nvim/lua/bufferline.lua:247>
```

### What did you expect to happen?

Should show a message: There is no buffer to close

### Config

```lua
return {
  {
    "akinsho/bufferline.nvim",
    event = "VeryLazy",
    keys = {
      { "<leader>bp", "<Cmd>BufferLineTogglePin<CR>",            desc = "Toggle pin" },
      { "<leader>bk", "<Cmd>BufferLineGroupClose ungrouped<CR>", desc = "Delete non-pinned buffers" },
      { "<leader>bn", ":e ~/.nvim-scratch<CR>",                  desc = "Open Scratch Buffer" },
      { "<C-s>",         "<cmd>BufferLineTogglePin<cr>",            desc = "Prev Buffer" },
[...]

PR #681, opened 2023-02-16
fix(group): check that group exists before operating on it

otavioschwanck:
Fixes https://github.com/akinsho/bufferline.nvim/issues/678

### DIFF
--- a/lua/bufferline/groups.lua
+++ b/lua/bufferline/groups.lua
@@ -301,6 +301,11 @@ function M.command(group_name, callback)
     function(list) return list.name == group_name end,
     state.components_by_group
   )
+
+  if not group then
+    return
+  end
+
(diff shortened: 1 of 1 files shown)


lua/bufferline/groups.lua
```diff
@@ -301,6 +301,11 @@ function M.command(group_name, callback)
     function(list) return list.name == group_name end,
     state.components_by_group
   )
+
+  if not(group) then
```
akinsho, line 305: Parentheses aren't required around conditional clauses in lua

akinsho (commented):
LGTM just one minor nit

lua/bufferline/groups.lua
otavioschwanck: @akinsho fixed!

Issue #679, opened 2023-02-15
[Bug]: Sometimes buffers tabs doesn't disappear when closing with CloseGroup


otavioschwanck:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

After calling BufferLineGroupClose ungrouped, the buffer names are still showing even that buffers are closed.  Error happened:

E21: Impossível fazer mudanças, 'modifiable' está desativado

### What did you expect to happen?

tab disappears

### Config

```lua
return {
  {
    "akinsho/bufferline.nvim",
    event = "VeryLazy",
    keys = {
      { "<leader>bp", "<Cmd>BufferLineTogglePin<CR>",            desc = "Toggle pin" },
      { "<leader>bk", "<Cmd>BufferLineGroupClose ungrouped<CR>", desc = "Delete non-pinned buffers" },
      { "<leader>bn", ":e ~/.nvim-scratch<CR>",                  desc = "Open Scratch Buffer" },
      { "<C-s>",         "<cmd>BufferLineTogglePin<cr>",            desc = "Prev Buffer" },
      { "[b",         "<cmd>BufferLineCyclePrev<cr>",            desc = "Prev Buffer" },
      { "]b",         "<cmd>BufferLineCycleNext<cr>",            desc = "Next Buffer" },
      { "H",         "<cmd>BufferLineCyclePrev<cr>",            desc = "Prev Buffer" },
      { "L",         "<cmd>BufferLineCycleNext<cr>",            desc = "Next Buffer" },
      { "[B",         "<cmd>BufferLineMovePrev<cr>",             desc = "Switch Prev Buffer" },
      { "]B",         "<cmd>BufferLineMoveNext<cr>",             desc = "Switch Next Buffer" },
      { "<C-h>",         "<cmd>BufferLineMovePrev<cr>",             desc = "Switch Prev Buffer" },
      { "<C-l>",         "<cmd>BufferLineMoveNext<cr>",             desc = "Switch Next Buffer" },
    },
    opts = {
      options = {
        diagnostics = "nvim_lsp",
        always_show_bufferline = true,
        offsets = {
          {
            filetype = "neo-tree",
            text = "Neo-tree",
            highlight = "Directory",
            text_align = "left",
          },
        },
      },
    },
  }
}
```

### Additional Information

Using :Telescope buffers fixes the problem.

### commit

013857c12340bc14e285727eaf7d0bfb9807c768

akinsho:
[...]

PR #682, opened 2023-02-16
fix(groups): refresh the UI when a group is closed

otavioschwanck:
fixes: https://github.com/akinsho/bufferline.nvim/issues/679


akinsho:
Thanks for the fix 👍🏿

PR #690, opened 2023-02-21
fix empty group

otavioschwanck:
image


lua/bufferline/utils/init.lua
```diff
@@ -107,7 +107,7 @@ end
 ---@param callback fun(item: `T`)
 ---@param matcher (fun(item: `T`):boolean)?
 function M.for_each(callback, list, matcher)
-  for _, item in ipairs(list) do
+  for _, item in ipairs(list or {}) do
```
akinsho, line 110: This function  should expect a list and be passed a list. The lack of a list here is a failing of the caller not of the function. I think the better solution would be to assert that a list is passed inside this function and find the place where the error is happening. Ie the place where nil is being passed and fix that not just defaulting it here

akinsho (changes requested):
@otavioschwanck Thanks for the PR. I've made a comment that I think we should find the source of the bug rather than just defaulting the nil value. That function should not be being passed nil in the first place, so we need to find where that is happening

lua/bufferline/utils/init.lua
otavioschwanck: The error is happening when i call:

`<Cmd>BufferLineGroupClose ungrouped`

with no ungroupeds

akinsho: But this is not the point in the code where an empty value is being passed to a for each function. What I'm saying is that we need to find where and when nil is being passed to the function and add a guard clause there.

otavioschwanck: @akinsho done

lua/bufferline/groups.lua
```diff
@@ -301,7 +301,7 @@ function M.command(group_name, callback)
     function(list) return list.name == group_name end,
     state.components_by_group
   )
-  utils.for_each(callback, group)
+  utils.for_each(callback, group or {})
```
akinsho, line 304: This function should not be being called at all if there is no matching group. If there is no group then it should do nothing. This is the point where it should be `if not group then return end`

akinsho: Are you planning another PR? my suggestion was just a small change above where you are doing this defaulting

Issue #692, opened 2023-02-21
[Bug]: After closing all groups, opening any buffer and try to pin, it throws an error


otavioschwanck:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

<img width="1288" alt="image" src="https://user-images.githubusercontent.com/9551316/220371555-5d8aa6fa-a4fd-4d96-8ae4-1da787521f58.png">

### What did you expect to happen?

Just pin

### Config

```lua
return {
  {
    "akinsho/bufferline.nvim", branch = 'dev',
    -- dir =  "~/Projetos/bufferline.nvim/",
    lazy = false,
    keys = {
      { "<leader>bp", "<Cmd>BufferLineTogglePin<CR>",                                                desc = "Toggle pin" },
      { "<leader>bk", "<Cmd>BufferLineGroupClose ungrouped<CR>",                                     desc = "Delete non-pinned buffers" },
      { "<leader>bK", "<Cmd>BufferLineGroupClose ungrouped<CR><Cmd>BufferLineGroupClose pinned<CR>", desc = "Delete all buffers" },
      { "<leader>bn", ":e ~/.nvim-scratch<CR>",                                                      desc = "Open Scratch Buffer" },
      { "<C-s>",      "<cmd>BufferLineTogglePin<cr>",                                                desc = "Prev Buffer" },
      { "[b",         "<cmd>BufferLineCyclePrev<cr>",                                                desc = "Prev Buffer" },
      { "]b",         "<cmd>BufferLineCycleNext<cr>",                                                desc = "Next Buffer" },
      { "H",          "<cmd>BufferLineCyclePrev<cr>",                                                desc = "Prev Buffer" },
      { "L",          "<cmd>BufferLineCycleNext<cr>",                                                desc = "Next Buffer" },
      { "[B",         "<cmd>BufferLineMovePrev<cr>",                                                 desc = "Switch Prev Buffer" },
      { "]B",         "<cmd>BufferLineMoveNext<cr>",                                                 desc = "Switch Next Buffer" },
      { "<C-h>",      "<cmd>BufferLineMovePrev<cr>",                                                 desc = "Switch Prev Buffer" },
      { "<C-l>",      "<cmd>BufferLineMoveNext<cr>",                                                 desc = "Switch Next Buffer" },
    },
    opts = {
      options = {
        diagnostics = "nvim_lsp",
        always_show_bufferline = true,
        offsets = {
          {
[...]

Issue #691, opened 2023-02-21
[Bug]: Pin buffers are not upgraded anymore (local.manual_groupings)


otavioschwanck:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

If i try to unpin some buffer, the g:BufferlinePinnedBuffers keeps the same.  If i restart the session, it continues on pinned.

### What did you expect to happen?

remove completely from pinned

### Config

```
return {
  {
    -- "akinsho/bufferline.nvim",
    dir = "~/Projetos/bufferline.nvim/",
    lazy = false,
    keys = {
      { "<leader>bp", "<Cmd>BufferLineTogglePin<CR>",                                                desc = "Toggle pin" },
      { "<leader>bk", "<Cmd>BufferLineGroupClose ungrouped<CR>",                                     desc = "Delete non-pinned buffers" },
      { "<leader>bK", "<Cmd>BufferLineGroupClose ungrouped<CR><Cmd>BufferLineGroupClose pinned<CR>", desc = "Delete all buffers" },
      { "<leader>bn", ":e ~/.nvim-scratch<CR>",                                                      desc = "Open Scratch Buffer" },
      { "<C-s>",      "<cmd>BufferLineTogglePin<cr>",                                                desc = "Prev Buffer" },
      { "[b",         "<cmd>BufferLineCyclePrev<cr>",                                                desc = "Prev Buffer" },
      { "]b",         "<cmd>BufferLineCycleNext<cr>",                                                desc = "Next Buffer" },
      { "H",          "<cmd>BufferLineCyclePrev<cr>",                                                desc = "Prev Buffer" },
      { "L",          "<cmd>BufferLineCycleNext<cr>",                                                desc = "Next Buffer" },
      { "[B",         "<cmd>BufferLineMovePrev<cr>",                                                 desc = "Switch Prev Buffer" },
      { "]B",         "<cmd>BufferLineMoveNext<cr>",                                                 desc = "Switch Next Buffer" },
      { "<C-h>",      "<cmd>BufferLineMovePrev<cr>",                                                 desc = "Switch Prev Buffer" },
      { "<C-l>",      "<cmd>BufferLineMoveNext<cr>",                                                 desc = "Switch Next Buffer" },
    },
    opts = {
      options = {
        diagnostics = "nvim_lsp",
        always_show_bufferline = true,
        offsets = {
          {
            filetype = "neo-tree",
            text = "Neo-tree",
            highlight = "Directory",
[...]

PR #693, opened 2023-02-21
fix(groups): ensure groups are persisted correctly

otavioschwanck:
Fixes https://github.com/akinsho/bufferline.nvim/issues/692 and https://github.com/akinsho/bufferline.nvim/issues/691

### DIFF
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -134,9 +134,7 @@ function M.group_action(name, action)
   assert(name, "A name must be passed to execute a group action")
   if action == "close" then
     groups.command(name, function(b) api.nvim_buf_delete(b.id, { force = true }) end)
-
     ui.refresh()
-
     groups.reset_manual_groupings()
   elseif action == "toggle" then
     groups.toggle_hidden(nil, name)
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -134,7 +134,10 @@ function M.group_action(name, action)
   assert(name, "A name must be passed to execute a group action")
   if action == "close" then
     groups.command(name, function(b) api.nvim_buf_delete(b.id, { force = true }) end)
+
     ui.refresh()
+
+    groups.reset_manual_groupings()
   elseif action == "toggle" then
     groups.toggle_hidden(nil, name)
     ui.refresh()
--- a/lua/bufferline/groups.lua
+++ b/lua/bufferline/groups.lua
@@ -187,8 +187,12 @@ local function persist_pinned_buffers()
   for buf, group in pairs(state.manual_groupings) do
     if group == PINNED_ID then table.insert(pinned, api.nvim_buf_get_name(buf)) end
   end
-  if #pinned == 0 then return end
-  vim.g[PINNED_KEY] = table.concat(pinned, ",")
+
+  if #pinned == 0 then
+    vim.g[PINNED_KEY] = ""
+  else
+    vim.g[PINNED_KEY] = table.concat(pinned, ",")
+  end
 end
 
 ---@param element TabElement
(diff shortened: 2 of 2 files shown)


lua/bufferline/groups.lua
```diff
@@ -169,7 +169,7 @@ builtin.pinned = Group:new({
 ----------------------------------------------------------------------------------------------------
 
 --- @type GroupState
-local state = {
+M.state = {
```
akinsho, line 172: This state should not be exposed. Instead you should create a method that you can export from here and use in the `bufferline.lua` file e.g. 
```lua
function M.reset_manual_groupings()
	state.manual_groupings = {}
end

-- In bufferline.lua use
groups.reset_manual_grouping()
```

lua/bufferline.lua
```diff
@@ -134,7 +135,12 @@ function M.group_action(name, action)
   assert(name, "A name must be passed to execute a group action")
   if action == "close" then
     groups.command(name, function(b) api.nvim_buf_delete(b.id, { force = true }) end)
+
+    vim.g[PINNED_KEY] = {}
+
     ui.refresh()
+
+    groups.state.manual_groupings = {}
```
akinsho, line 143: Similarly the pinned key is only relevant to the groupings and should not be leaked outside that module because it shares unnecessary implementation details. You can add this line to helper function I suggested in `groups.lua`

lua/bufferline.lua
```diff
@@ -230,11 +236,11 @@ local function setup_commands()
   cmd("BufferLinePick", function() M.pick_buffer() end, {})
   cmd("BufferLinePickClose", function() M.close_buffer_with_pick() end, {})
   cmd("BufferLineCycleNext", function() M.cycle(1) end, {})
-  cmd("BufferLineCyclePrev", function() M.cycle(-1) end, {})
+  cmd("BufferLineCyclePrev", function() M.cycle( -1) end, {})
```
akinsho, line 239: Would be good to avoid unnecessary changes like this so this PR is easier to review

lua/bufferline/groups.lua
```diff
@@ -329,7 +333,10 @@ function M.is_pinned(element) return get_manual_group(element) == PINNED_ID end
 ---@param element TabElement?
 function M.add_to_group(group_name, element)
   local group = group_by_name(group_name)
-  if group and element then set_manual_group(element.id, group.id) end
+  if group and element then
+    set_manual_group(element.id, group.id)
+    persist_pinned_buffers()
```
akinsho, line 338: A PR I made on `dev` recently removed this line and added it in elsewhere. Is this re-added intentionally?

akinsho (changes requested):
@otavioschwanck thanks for the PR. I mentioned in my review that I don't want details that are specific to the groups module being outside of that space since it causes the main module to depend on the implementation of `groups.lua` so I suggest creating a helper and exporting that, rather than exposing private state.

lua/bufferline/groups.lua
otavioschwanck: Yes, it has to be in it to work properly.

lua/bufferline.lua
otavioschwanck: Just lua lint, sorry

lua/bufferline.lua
otavioschwanck: Done

lua/bufferline/groups.lua
otavioschwanck: Done

lua/bufferline.lua
```diff
@@ -134,7 +135,12 @@ function M.group_action(name, action)
   assert(name, "A name must be passed to execute a group action")
   if action == "close" then
     groups.command(name, function(b) api.nvim_buf_delete(b.id, { force = true }) end)
+
+    vim.g[PINNED_KEY] = {}
```
akinsho, line 139: It would also be good to not expose this here, can you move this into the `reset_manual_groups` function. We basically now have the `PINNED_KEY` maintained as a string in two places and they can fall out of sync

lua/bufferline.lua
akinsho: Can you revert these changes, since stylua will not like them and will try to fix it anyway

akinsho (changes requested):
Getting there, just a couple more comments. Would really like to see the Pinned key not brought into the main bufferline module

otavioschwanck:
Changes done!

akinsho:
LGTM 👍🏿 thanks

Issue #695, opened 2023-02-22
[Bug]: BufferLineGroupClose ungrouped is removing the pin of pinned tabs


otavioschwanck:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

[Bug]: BufferLineGroupClose ungrouped is removing the pin of pinned tabs

### What did you expect to happen?

The pins should be keeped

### Config

```
return {
  {
    "akinsho/bufferline.nvim", branch = 'dev',
    lazy = false,
    keys = {
      { "<leader>bp", "<Cmd>BufferLineTogglePin<CR>",            desc = "Toggle pin" },
      { "<leader>bk", "<Cmd>BufferLineGroupClose ungrouped<CR>", desc = "Delete non-pinned buffers" },
      { "<leader>bK", "<Cmd>bufdo bwipeout<CR>",                 desc = "Kill all buffers" },
      { "<leader>bn", ":e ~/.nvim-scratch<CR>",                  desc = "Open Scratch Buffer" },
      { "<C-s>",      "<cmd>BufferLineTogglePin<cr>",            desc = "Prev Buffer" },
      { "[b",         "<cmd>BufferLineCyclePrev<cr>",            desc = "Prev Buffer" },
      { "]b",         "<cmd>BufferLineCycleNext<cr>",            desc = "Next Buffer" },
      { "H",          "<cmd>BufferLineCyclePrev<cr>",            desc = "Prev Buffer" },
      { "L",          "<cmd>BufferLineCycleNext<cr>",            desc = "Next Buffer" },
      { "[B",         "<cmd>BufferLineMovePrev<cr>",             desc = "Switch Prev Buffer" },
      { "]B",         "<cmd>BufferLineMoveNext<cr>",             desc = "Switch Next Buffer" },
      { "<C-h>",      "<cmd>BufferLineMovePrev<cr>",             desc = "Switch Prev Buffer" },
      { "<C-l>",      "<cmd>BufferLineMoveNext<cr>",             desc = "Switch Next Buffer" },
    },
    opts = {
      options = {
        diagnostics = "nvim_lsp",
        always_show_bufferline = false,
        persist_buffer_sort = true,
        offsets = {
          {
            filetype = "neo-tree",
            text = "Neo-tree",
            highlight = "Directory",
            text_align = "left",
          },
        },
      },
    },
  }
}
```

### commit

894d0e80583cacd27facdb6676a6b08764662537

akinsho:
[...]

PR #696, opened 2023-02-22
fix(groups): only reset the pinned group if it closed

otavioschwanck:
fixes: https://github.com/akinsho/bufferline.nvim/issues/695

### DIFF
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -135,7 +135,10 @@ function M.group_action(name, action)
   if action == "close" then
     groups.command(name, function(b) api.nvim_buf_delete(b.id, { force = true }) end)
     ui.refresh()
-    groups.reset_manual_groupings()
+
+    if name == groups.pinned_name then
+      groups.reset_manual_groupings()
+    end
   elseif action == "toggle" then
     groups.toggle_hidden(nil, name)
     ui.refresh()
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -136,9 +136,7 @@ function M.group_action(name, action)
     groups.command(name, function(b) api.nvim_buf_delete(b.id, { force = true }) end)
     ui.refresh()
 
-    if name == groups.pinned_name then
-      groups.reset_manual_groupings()
-    end
+    groups.reset_manual_groupings(name)
   elseif action == "toggle" then
     groups.toggle_hidden(nil, name)
     ui.refresh()
--- a/lua/bufferline/groups.lua
+++ b/lua/bufferline/groups.lua
@@ -468,9 +468,9 @@ function M.reset_manual_groupings(name)
     vim.g[PINNED_KEY] = {}
   end
 
-  for _, group_id in pairs(state.manual_groupings) do
+  for buf, group_id in pairs(state.manual_groupings) do
     if group_id == name then
-      state.manual_groupings[group_id] = nil
+      state.manual_groupings[buf] = nil
     end
   end
 end
--- a/lua/bufferline/groups.lua
+++ b/lua/bufferline/groups.lua
@@ -466,7 +466,12 @@ end
 function M.reset_manual_groupings(name)
   if name == PINNED_NAME then
     vim.g[PINNED_KEY] = {}
-    state.manual_groupings = {}
+  end
+
+  for _, group_id in pairs(state.manual_groupings) do
+    if group_id == name then
+      state.manual_groupings[group_id] = nil
+    end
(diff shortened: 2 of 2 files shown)


lua/bufferline/groups.lua
```diff
@@ -517,4 +517,6 @@ if utils.is_test() then
   M.set_manual_group = set_manual_group
 end
 
+M.pinned_name = PINNED_NAME
```
akinsho, line 520: Rather than exposing this, instead this logic should just be done inside of the `reset_manual_groupings`

lua/bufferline.lua
```diff
@@ -135,7 +135,10 @@ function M.group_action(name, action)
   if action == "close" then
     groups.command(name, function(b) api.nvim_buf_delete(b.id, { force = true }) end)
     ui.refresh()
-    groups.reset_manual_groupings()
+
+    if name == groups.pinned_name then
```
akinsho, line 139: This logic should not be here, it breaks the encapsulation of that module. In general things should stay in the modules unless absolutely necessary for example now that the name is exposed `bufferline` in *general* now depends on the name, but that could be an implementation to change later so it shouldn't be exposed

akinsho (changes requested):
Thanks again for the PR. I've made some comments. Essentially I think this logic should moved entirely into the `groups.lua` function. I think that when deciding whether or not to reset manual groups *there* is where the check should happen

lua/bufferline.lua
otavioschwanck: Ok!

otavioschwanck:
Changes done!

lua/bufferline/groups.lua
```diff
@@ -462,9 +462,12 @@ local function sort_by_groups(components)
...
 
 --- Resets manual mappings.
-function M.reset_manual_groupings()
-  vim.g[PINNED_KEY] = {}
-  state.manual_groupings = {}
+---@param name string Group name
+function M.reset_manual_groupings(name)
+  if name == PINNED_NAME then
+    vim.g[PINNED_KEY] = {}
+    state.manual_groupings = {}
```
akinsho, line 469: This shouldn't clear all groupings, I didn't realise what was happening sooner but a user can manually add a buffer to a group e.g.

```lua
local manual_groupings = {
  5 = "custom_docs_group",
  11 = "test_files",
  17 = "pinned",
  3 = "pinned"
}
```

the way your logic is written currently, then if I had added custom groups then this change would break my groups as well as clear the pinned ones.

Instead the logic should be
```lua
   if name == PINNED_NAME then
     vim.g[PINNED_KEY] = {}
   end
   for _, group_id in pairs(state.manual_groupings)
     if group_id == name then
		state.manual_groupings[group_id] = nil
     end
   end
```

This way this function can handle other types of groups without just clearing everything

akinsho (commented):
@otavioschwanck now I look at it I think I didn't understand some of these changes.

So `manual_groupings` isn't only for `pinned` buffers it is a general state holder for any buffer to be assigned to a group by a user. A user can choose to manually add a buffer to a group so clearing all of it is actually not correct/causes bugs with the other functionality.

I've left an explanation of the change needed it's a small code change

lua/bufferline/groups.lua
otavioschwanck: > ```lua
>    if name == PINNED_NAME then
>      vim.g[PINNED_KEY] = {}
>    end
>    for _, group_id in pairs(state.manual_groupings)
>      if group_id == name then
> 		state.manual_groupings[group_id] = nil
>      end
>    end
> ```

done!

lua/bufferline/groups.lua
```diff
@@ -466,7 +466,12 @@ end
 function M.reset_manual_groupings(name)
   if name == PINNED_NAME then
     vim.g[PINNED_KEY] = {}
-    state.manual_groupings = {}
+  end
+
+  for _, group_id in pairs(state.manual_groupings) do
+    if group_id == name then
+      state.manual_groupings[group_id] = nil
```
akinsho, line 473: Sorry I didn't check the logic properly as it's been a long day you need to use the key which is the buffer number to delete the item

```lua
for buf, group_id ... do
	state.manual_groupings[buf] = nil
```

Please check the code carefully as well 🙏🏿

otavioschwanck: oh shit, lucky you see it!

changes done!

akinsho (approved):
LGTM 👍🏿

Issue #697, opened 2023-02-22
[Bug]: Closed buffer stays pinned if open again


otavioschwanck:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

It stays pinned

### What did you expect to happen?

The pin should be removed

### Config

I think you already know hahah

### Additional Information

Maybe we should have a close action for buffers (that handles the pin removal).  If you want, i can do it, just say the word

### commit

dev latest

akinsho:
@otavioschwanck tbh I'm not sure about this one. I kind of like that pinned buffers remain pinned even if they are closed since I tend to pin buffers that are quite important so whenever I re-open them I want them to still be pinned. I consider a feature. I think if people don't want this behaviour they should be careful to unpin the buffer.

otavioschwanck:
> @otavioschwanck tbh I'm not sure about this one. I kind of like that pinned buffers remain pinned even if they are closed since I tend to pin buffers that are quite important so whenever I re-open them I want them to still be pinned. I consider a feature. I think if people don't want this behaviour they should be careful to unpin the buffer.

The problem is: with bwipeout all buffers, it throws an error if i try to pin again:

<img width="1251" alt="image" src="https://user-images.githubusercontent.com/9551316/220754604-ab4ca7d8-650d-4ba2-b524-27f25088f5b3.png">

(Closed with :bufdo bwipeout)

akinsho:
@otavioschwanck using `bwipeout` is a bad idea unless you have a very good very specific reason for that. From the help docs 👇🏿 

>  		Like |:bdelete|, but really delete the buffer.  Everything
> 		related to the buffer is lost.  All marks in this buffer
> 		become invalid, option settings are lost, etc.  Don't use this
> 		unless you know what you are doing. Examples:  

By using `bdelete` you free the buffer number so it will now be completely invalid an error. In this case I think there should be error handling but the problem is mainly with the usage of `bwipeout` which I think is wrong.

> 
> And other thing:
> 
> When i close a pinned buffer and open two months later, it will be pinned, when i close a pinned buffer

I think this is fine, I think users should unpin a buffer if they don't want it pinned I don't think closing should be assumed to be the same as unpinning. It's not a question of complexity and I definitely don't want new commands, this plugin is already big enough.

akinsho:
Merged #698 as a way to do this... for now, not sure if the API will change though

PR #698, opened 2023-02-23
Feature/close and unpin

otavioschwanck:
Fixes https://github.com/akinsho/bufferline.nvim/issues/697

PS:
i didn't create a public commands.


akinsho:
Really annoyingly this gets autoclosed when I merge `dev` into `main` because that branch keeps getting autoclosed. I'm kind of fine with this functionality but a bit worried about all these very special purpose functions like remove manual grouping. Not sure if having them means something needs refactoring (Just thinking out loud)

Issue #569, opened 2022-09-26
[Bug]: Wrong separator for tabs


Nik-i-Net:
### What happened?

When using "mode = buffers", tab sepatators are wrong.

### What did you expect to happen?

Correct separators or the ability to put a standard separator for tabs.

### Config

![Screenshot from 2022-09-26 14-22-12](https://user-images.githubusercontent.com/91474300/192265587-90e6f36a-102f-4a50-a12c-4b8855533e7f.png)

### Additional Information

This behavior on any color scheme.

akinsho:
@Nik-i-Net this issue is invalid as there is no relationship between separator style and mode. You have set the separator style to slant, and it appears as slant which is what it is supposed to do. If you still believe something is wrong please clarify *very clearly what* and provide a *minimal init.lua* (not a screenshot) that I can use to reproduce any issue

Luis-Licea:
I am having the same issue. The slanted characters are supposed to occur on both sides of the tab, but only occur on one. This does not depend on the color scheme, as Nik mentioned. Here are some examples.
![image](https://user-images.githubusercontent.com/48164786/194729826-25ab9f04-71db-4f45-baec-423febf21661.png)
![image](https://user-images.githubusercontent.com/48164786/194729917-cfcb6bd3-3f8c-4a46-8ec9-654f8ea452d2.png)

akinsho:
@Luis-Licea the minimal tab indicators aren't configurable with custom separators, they aren't expected to have slanted sides at all, I'm not really sure how you're achieving that image in the screenshot but it definitely isn't a supported feature so I'd say this isn't a bug since I never built it to do what you're trying to do. Only the actual full buffer tabs can have custom separators

knmac:
I think the problem here is that the "tab tabs" are rendered with only one separator. This is seen in the function ` render(tabpage, is_active, style, highlights)` in the file `tabpages.lua`:

Original code:
```lua
local function render(tabpage, is_active, style, highlights)
  local h = highlights
  local hl = is_active and h.tab_selected.hl_group or h.tab.hl_group
  local separator_hl = is_active and h.tab_separator_selected.hl_group or h.tab_separator.hl_group
  local chars = constants.sep_chars[style] or constants.sep_chars.thin
  local separator_component = chars[2]
  local name = padding .. padding .. tabpage.tabnr .. padding
  return {
    { highlight = hl, text = name, attr = { prefix = tab_click_component(tabpage.tabnr) } },
[...]

PR #714, opened 2023-03-22
Fix: tabs do not use seperators correctly

matthis-k:
This is the PR for #569.
This works perfectly for `slant` and `slope`, but I am not sure what the behavior for `thick` and `thin` is supposed to be.
If you want the behavior to be unchanged I (or you) can change it this to this, which should only change the `slant` and `slope` and custom styles:
```
  local single_char = require("bufferline.config").options.separator_style == "thick"
    or require("bufferline.config").options.separator_style == "thin"
  return {
    single_char and {} or { highlight = separator_hl, text = chars[2] },
    {
      highlight = hl,
      text = name,
      attr = { prefix = tab_click_component(tabpage.tabnr) },
    },
    { highlight = separator_hl, text = chars[single_char and 2 or 1] },

```

### DIFF
--- a/lua/bufferline/tabpages.lua
+++ b/lua/bufferline/tabpages.lua
@@ -29,11 +29,11 @@ local function render(tabpage, is_active, style, highlights)
   local hl = is_active and h.tab_selected.hl_group or h.tab.hl_group
   local separator_hl = is_active and h.tab_separator_selected.hl_group or h.tab_separator.hl_group
   local chars = constants.sep_chars[style] or constants.sep_chars.thin
-  local separator_component = chars[2]
-  local name = padding .. padding .. tabpage.tabnr .. padding
+  local name = padding .. tabpage.tabnr .. padding
   return {
+    { highlight = separator_hl, text = chars[2] },
     { highlight = hl, text = name, attr = { prefix = tab_click_component(tabpage.tabnr) } },
-    { highlight = separator_hl, text = separator_component },
+    { highlight = separator_hl, text = chars[1] },
   }
 end
 


akinsho:
Hi @matthis-k thanks for the PR, whilst it would be good to have this fixed I definitely would rather not have *any regressions* with it since this will definitely lead to 1 issue being closed and 3 opened to replace it.

Currently I see 

Screenshot 2023-03-24 at 10 53 23

When using your PR i.e. there are strange gaps in between characters which has never historically been the case. I've also never been able to reproduce the parent issue here and am curious what terminal emulator causes this

matthis-k:
I think this is a mistake on my part, I think the issue is that the `slanted` style wants to render `slant-tab-slant` and `thin` only wants `tab-char`. I'll fo and fix it.

lua/bufferline/tabpages.lua
```diff
@@ -29,12 +29,19 @@ local function render(tabpage, is_active, style, highlights)
   local hl = is_active and h.tab_selected.hl_group or h.tab.hl_group
   local separator_hl = is_active and h.tab_separator_selected.hl_group or h.tab_separator.hl_group
   local chars = constants.sep_chars[style] or constants.sep_chars.thin
-  local separator_component = chars[2]
-  local name = padding .. padding .. tabpage.tabnr .. padding
+  local name = padding .. tabpage.tabnr .. padding
+  local swapped_chars = require("bufferline.config").options.separator_style == "thick"
```
akinsho, line 33: This import should match those above rather than be inline like this i.e.
```lua
---@module "bufferline.config"
local config = lazy.require("bufferline.config")
```
Although it appears **the style is already being passed into this function as an argument** so I don't thing you need to even be requiring the config at all.

akinsho: This check could also be done a little more tersely with something like
```lua
local style = config.options.separator_style
local swapped_chars = ({thick = {1, 2}, thin = {1, 2}})[style] or {2, 1}
return {
 { highlight = separator_hl, text = swapped_chars[1] },
 -- ...
 { highlight = separator_hl, text = swapped_chars[2] },
```

akinsho (changes requested):
@matthis-k thanks for tweaking it, seems like it looks better now. I've made some code style comments since I think the require is unnecessary and the conditional can be terser still.

akinsho:
It's also worth clarifying that it isn't a bug that the sloped style and the minimal tabs don't look the same, they were never intended to. This is a good change but just for clarity the minimal tabs were always initially intended to use the thin style regardless of if a user set sloped or anything like that

matthis-k:
> It's also worth clarifying that it isn't a bug that the sloped style and the minimal tabs don't look the same, they were never intended to. This is a good change but just for clarity the minimal tabs were always initially intended to use the thin style regardless of if a user set sloped or anything like that

Oh I see.  I thought it was spposed to work, as you would still use the chars from the style here: `local chars = constants.sep_chars[style] or constants.sep_chars.thin`

PR #726, opened 2023-04-14
fix(ui): protect against nil duplicate_prefix

bradsherman:
Today I started getting the following error:

```
E5108: Error executing lua ...al/share/nvim/lazy/bufferline.nvim/lua/bufferline/ui.lua:134: Segments must be a list
stack traceback:
        [C]: in function 'assert'
        ...al/share/nvim/lazy/bufferline.nvim/lua/bufferline/ui.lua:134: in function 'get_component_size'
        ...al/share/nvim/lazy/bufferline.nvim/lua/bufferline/ui.lua:514: in function <...al/share/nvim/lazy/bufferline.nvim/lua/bufferline/ui.lua:487>
        vim/shared.lua: in function 'get_components'
        ...local/share/nvim/lazy/bufferline.nvim/lua/bufferline.lua:85: in function <...local/share/nvim/lazy/bufferline.nvim/lua/bufferline.lua:81>
```

I did a little digging and realized it was due to `duplicate_prefix` being nil [here](https://github.com/akinsho/bufferline.nvim/blob/main/lua/bufferline/ui.lua#L504) which caused the table to begin with index 2. This causes the the [`tbl_islist`](https://github.com/akinsho/bufferline.nvim/blob/main/lua/bufferline/ui.lua#L132) assert to fail.


bradsherman:
@akinsho This is a workaround I'm using for the time being, but let me know if there's a more correct/idiomatic solution.

hood:
Chiming in to say I'm experienced the same problem here, opened an issue for it: #727

bradsherman:
I'm building neovim from master

```
NVIM v0.10.0-dev-1431+g002b80cb5
Build type: Release
LuaJIT 2.1.0-beta3
```

Here's my `bufferline` config:
```lua
local status_ok, bufferline = pcall(require, "bufferline")
if not status_ok then
    return
end

bufferline.setup({
    options = {
        numbers = "none",
        close_command = "Bdelete! %d",
        right_mouse_command = "Bdelete! %d",
        left_mouse_command = "buffer %d",
        middle_mouse_command = nil,
        indicator = {
            icon = "▎",
            style = "icon",
        },
        buffer_close_icon = "",
        modified_icon = "●",
        close_icon = "",
        left_trunc_marker = "",
        right_trunc_marker = "",
        max_name_length = 18,
        max_prefix_length = 15,
        tab_size = 18,
        diagnostics = "nvim_lsp",
        diagnostics_update_in_insert = false,
        offsets = {
            {
                filetype = "NvimTree",
                text = function()
                    return vim.fn.getcwd()
                end,
                highlight = "Directory",
                text_align = "left",
                padding = 1,
            },
            {
                filetype = "neo-tree",
                text = function()
                    return vim.fn.getcwd()
                end,
                highlight = "Directory",
                text_align = "left",
                padding = 1,
            },
        },
        color_icons = true,
        show_buffer_icons = true,
        show_close_icon = true,
        show_tab_indicators = true,
        persist_buffer_sort = true,
        separator_style = "padded_slant",
        enforce_regular_tabs = true,
    },
})
```

akinsho:
Ah right I've had a look now and this is due to a breaking change in nvim HEAD. I personally don't track head super closely so things can and will break there. Anyway the change seems to be that the `vim.tbl_islist` has been made stricter and no longer works for this check. I think the best solution is to use the newly added/rename `vim.tbl_isarray`.

What I suggest is to go to where `tbl_islist` is used and check if `tbl_isarray` exists first if it does use that otherwise use is list.

e.g.
```lua
local is_list = vim.tbl_isarray or vim.tbl_islist
if is_list(values) then
 ---
end
```

lua/bufferline/utils/init.lua
```diff
@@ -242,4 +242,10 @@ function M.is_truthy(value)
     and value ~= "nil"
 end
 
+--- Determine which list-check function to use
```
akinsho, line 245: @bradsherman could please add add a `TODO: deprecate this in nvim-0.11 or use strict lists`

bradsherman:
@akinsho sweet, I addressed your comment. Thanks for taking a look quickly!

akinsho:
Oh I think the sorter test broke 😮‍💨

bradsherman:
@akinsho sorry about that - I see the problem I'll submit a fix in a few minutes

PR #728, opened 2023-04-14
fix(utils): fix utils.is_list

bradsherman:
https://github.com/akinsho/bufferline.nvim/pull/726 Introduced a bug because `utils.is_list` need to be executed in order to return a function that will return a boolean if the object is a list. This broke the tests, now it properly returns a `boolean`.

I ran the tests with the following command I got from the ci workflow yaml.
```sh
nvim --headless -u tests/minimal_init.lua -c "PlenaryBustedDirectory tests/ {minimal_init = 'tests/minimal_init.lua', sequential = true}"
```

I also ran the formatter with:
```sh
stylua --config-path=stylua.toml lua/
```

### DIFF
--- a/lua/bufferline/utils/init.lua
+++ b/lua/bufferline/utils/init.lua
@@ -242,9 +242,12 @@ function M.is_truthy(value)
     and value ~= "nil"
 end
 
+local mk_is_list = function() return vim.tbl_isarray or vim.tbl_islist end
+local _is_list = mk_is_list()
+
 -- TODO: deprecate this in nvim-0.11 or use strict lists
 --- Determine which list-check function to use
----@return function
-function M.is_list() return vim.tbl_isarray or vim.tbl_islist end
+---@return boolean
+function M.is_list(l) return _is_list(l) end
 
 return M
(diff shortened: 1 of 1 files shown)


bradsherman:
@akinsho this should fix the problem introduced by the previous PR

lua/bufferline/utils/init.lua
```diff
@@ -242,9 +242,12 @@ function M.is_truthy(value)
     and value ~= "nil"
 end
 
+local mk_is_list = function() return vim.tbl_isarray or vim.tbl_islist end
+local _is_list = mk_is_list()
```
akinsho, line 246: I think this change can be inline i.e. and simplified i.e.

```lua
M.is_list = vim.tbl_isarray or vim.tbl_islist
```
I believe this line could replaace all of the other logic

akinsho (changes requested):
Thanks for tracking down the bug and raising the PR @bradsherman I've just left a comment as I think this logic can be simplified greatly

bradsherman:
Sorry I was away from my computer this weekend. That's much nicer - thanks for taking it over the finish line!

PR #731, opened 2023-04-18
Use new isarray function

Ursescu:
On dev neovim branch there is a change (https://github.com/neovim/neovim/commit/7caf0eaf) that modifies the islist function. Modifying bufferline ui to use isarray function (to keep the same functionality)


Ursescu:
Not sure if (bufferline) dev branch is active or not though ..

Ursescu:
It seems this is already addressed by #726, sorry for the noise.

PR #738, opened 2023-04-23
feat(diag): sane fallback to underline color

antoineco:
**tl;dr** To support colorschemes where LSP diagnostics are highlighted with colored underlines instead of foreground text (see screenshot below), check whether the `special` attribute exists before using the fallback group.

image

Result, here with an error diagnostic:

image

---

My main motivation is to allow a sensible fallback to the <ins>underline</ins> color in case any of the `DiagnosticError`, `DiagnosticWarn`, `DiagnosticHint` or `DiagnosticInfo` highlight groups do not have a foreground color set.

In many IDEs, it is a [common practice](https://user-images.githubusercontent.com/7727148/38758132-7284e58a-3f24-11e8-8e39-627e6916e9ec.PNG) to highlight diagnostics by underlining the context, instead of changing its foreground color. Due to Vim's legacy of supporting terminals without capabilities such as _italics_, **bold** or <ins>underline</ins>, this has never become a widespread pattern in our favourite text editor. Nevertheless, some (fairly popular) color schemes — such as [everforest](https://github.com/sainnhe/everforest), [gruvbox-material](https://github.com/sainnhe/gruvbox-material) or [sonokai](https://github.com/sainnhe/sonokai) — do apply underline styles to these highlight groups, and right now bufferline can't handle this practice very well (none of the colors below match the colorscheme's):

_error_
error

_warn_
warn

_hint_
hint

_info_
info

I believe that it is sensible and unintrusive to check for the existence of the `special` attribute during the initialization of `error_fg`, `warning_fg`, etc. before moving on to the final fallback.

Fixes https://github.com/sainnhe/everforest/issues/121

### DIFF
--- a/lua/bufferline/colors.lua
+++ b/lua/bufferline/colors.lua
@@ -48,7 +48,7 @@ end
 local new_hl_api = api.nvim_get_hl ~= nil
 local get_hl = function(name, use_cterm)
   if new_hl_api then
-    local hl = api.nvim_get_hl(0, { name = name })
+    local hl = api.nvim_get_hl(0, { name = name, link = false })
     if use_cterm then
       hl.fg, hl.bg = hl.ctermfg, hl.ctermbg
     end
--- a/lua/bufferline/colors.lua
+++ b/lua/bufferline/colors.lua
@@ -58,6 +58,14 @@ local get_hl = function(name, use_cterm)
   return api.nvim_get_hl_by_name(name, not use_cterm)
 end
 
+-- Map of nvim_get_hl() highlight attributes (new API) to
+-- nvim_get_hl_by_name() highlight attributes (old API).
+local hl_color_attrs = {
+  fg = "foreground",
+  bg = "background",
+  sp = "special",
+}
+
 -- parses the gui hex color code (or cterm color number) from the given hl_name
 --   color number (0-255) is returned if cterm is set to true in opts
 --   if unable to parse, uses the fallback value
@@ -68,11 +76,13 @@ function M.get_color(opts)
     opts.name, opts.attribute, opts.fallback, opts.not_match, opts.cterm
   -- translate from internal part to hl part
   assert(
-    attribute == "fg" or attribute == "bg",
-    fmt('attribute for %s should be one of "fg" or "bg", "%s" was passed in ', name, attribute)
+    hl_color_attrs[attribute],
+    fmt('unsupported attribute %s for %s, should be one of %s', attribute, name, table.concat(vim.tbl_keys(hl_color_attrs), ", "))
   )
   -- TODO: remove when 0.9 is stable
-  if not new_hl_api then attribute = attribute == "fg" and "foreground" or "background" end
+  if not new_hl_api then
+    attribute = hl_color_attrs[attribute]
+  end
 
   -- try and get hl from name
   local success, hl = pcall(get_hl, name, cterm)
--- a/lua/bufferline/config.lua
+++ b/lua/bufferline/config.lua
@@ -187,27 +187,55 @@ local function derive_colors(preset)
   local string_fg = hex({ name = "String", attribute = "fg" })
 
   local error_fg = hex({
-    name = "DiagnosticError",
+    name = "DiagnosticError", -- diagnostic with text highlight
     attribute = "fg",
-    fallback = { name = "Error", attribute = "fg" },
+    fallback = {
+      name = "DiagnosticError", -- diagnostic with underline highlight
+      attribute = "sp",
+      fallback = {
+        name = "Error",
+        attribute = "fg",
+      },
+    },
   })
 
   local warning_fg = hex({
     name = "DiagnosticWarn",
     attribute = "fg",
-    fallback = { name = "WarningMsg", attribute = "fg" },
+    fallback = {
+      name = "DiagnosticWarn",
+      attribute = "sp",
+      fallback = {
+        name = "WarningMsg",
+        attribute = "fg",
+      },
+    },
   })
 
   local info_fg = hex({
     name = "DiagnosticInfo",
     attribute = "fg",
-    fallback = { name = "Normal", attribute = "fg" },
+    fallback = {
+      name = "DiagnosticInfo",
+      attribute = "sp",
+      fallback = {
+        name = "Normal",
+        attribute = "fg",
+      },
+    },
   })
 
   local hint_fg = hex({
     name = "DiagnosticHint",
     attribute = "fg",
-    fallback = { name = "Directory", attribute = "fg" },
+    fallback = {
+      name = "DiagnosticHint",
+      attribute = "sp",
+      fallback = {
+        name = "Directory",
+        attribute = "fg",
+      },
+    },
   })
 
   local tabline_sel_bg = hex({


lua/bufferline/colors.lua
```diff
@@ -48,7 +48,7 @@ end
 local new_hl_api = api.nvim_get_hl ~= nil
 local get_hl = function(name, use_cterm)
   if new_hl_api then
-    local hl = api.nvim_get_hl(0, { name = name })
+    local hl = api.nvim_get_hl(0, { name = name, link = false })
```
antoineco, line 51: We need to follow links here to access the actual colors in case the highlight group is linked (unintuitively, this is done with `link = false`).

Neovim enforces a defaut foreground color for all core diagnostic highlight groups, and this color is wrongly returned when the highlight group is linked:

image

With `link = true` (the default):

```lua
{
  ctermfg = 1,
  fg = 16711680,
  link = "ErrorText"
}
```

With `link = false`:

```lua
{
  cterm = {
    undercurl = true
  },
  sp = 15105664,
  undercurl = true
}
```

akinsho: 🤔 Whilst understandable I really worry about how this will affect all the other users. There seems like a very high chance that this is going to lead to a lot of hard to resolve issues because a lot of users will see their bufferline highlights change in a way they won't understand

lua/bufferline/colors.lua
```diff
@@ -68,11 +68,18 @@ function M.get_color(opts)
...
   assert(
-    attribute == "fg" or attribute == "bg",
-    fmt('attribute for %s should be one of "fg" or "bg", "%s" was passed in ', name, attribute)
+    attribute == "fg" or attribute == "bg" or attribute == "sp",
+    fmt('attribute for %s should be one of "fg", "bg" or "sp", "%s" was passed in ', name, attribute)
   )
   -- TODO: remove when 0.9 is stable
-  if not new_hl_api then attribute = attribute == "fg" and "foreground" or "background" end
+  if not new_hl_api then
+    local old_to_new_attr = {
```
akinsho, line 76:
```suggestion
    attribute = ({
    	fg = "foreground",
    	bg = "background",
    	sp = "special"
    })[attribute]
```

lua/bufferline/colors.lua
```diff
@@ -68,11 +68,18 @@ function M.get_color(opts)
     opts.name, opts.attribute, opts.fallback, opts.not_match, opts.cterm
   -- translate from internal part to hl part
   assert(
-    attribute == "fg" or attribute == "bg",
-    fmt('attribute for %s should be one of "fg" or "bg", "%s" was passed in ', name, attribute)
+    attribute == "fg" or attribute == "bg" or attribute == "sp",
```
akinsho, line 71: At this point in wonder if the mapping of long name to abbreviation e.g. `fg = "foreground"` shouldn't be pulled out into a separate table and then this check could be if not `mapping_table[attribute]` and then also re-used later for the remapping

lua/bufferline/colors.lua
antoineco: I would argue that the current behavior is more error prone. When the highlight group is linked, Neovim displays the linked color in its UI, but bufferline picks Neovim's default instead, so `DiagnosticError` is effectively 2 different colors depending on how the plugin uses it.

antoineco: Since the behavior seems specific to those highlight groups, I don't mine moving this specific change to its own PR.

akinsho: So if I understand correctly the link is taking precedence over it's overrides? rather than the override showing first? I guess I wonder with this if a colour is strictly a link to some other color e.g. `nvim_set_hl(0, 'MyGroup', {link = 'OtherGroup'})` how will this change here behave with `link = false` what kind of result would we get for `MyGroup` because if it won't return `OtherGroup` then I read this as a breaking change but also a problem since this is desirable behaviour that we are no longer using.

The second issue/question is that what does the link persist if not the main object of a highlight group e.g. if I was creating a colorscheme I would completely replace my highlight group with the default e.g. `nvim_set_hl(0, 'DiagnosticInfo', {...my_opts..})` this would *replace* the highlight group and it's link so when the user tries to resolve `DiagnosticInfo` I wouldn't expect it to show a link anymore since my colorscheme should have removed the link i.e. why is this a problem for your theme? shouldn't it be completely overriding the theme?

lua/bufferline/colors.lua
```diff
@@ -68,11 +68,17 @@ function M.get_color(opts)
...
+    fmt('attribute for %s should be one of "fg", "bg" or "sp", "%s" was passed in ', name, attribute)
   )
   -- TODO: remove when 0.9 is stable
-  if not new_hl_api then attribute = attribute == "fg" and "foreground" or "background" end
+  if not new_hl_api then
+    attribute = ({
+      fg = "foreground",
+      bg = "background",
+      sp = "special",
+    }).attribute
```
akinsho, line 80: `}).attribute` looks wrong, I think this should be `})[attribute]`

lua/bufferline/colors.lua
antoineco: Done in [8e6c92a4..7836b304](https://github.com/akinsho/bufferline.nvim/compare/8e6c92a457c3aca12a8e43da6e2aacff25612f29..7836b304bfc6eb26d26d9714a6fa42af64d2734e)

lua/bufferline/colors.lua
```diff
@@ -58,6 +58,27 @@ local get_hl = function(name, use_cterm)
...
+-- Returns a pretty-printed list of supported highlight attributes.
+---@return string
+local function supported_hl_color_attr()
+  local attrs = ''
+  local i = 0
+  for attr, _ in pairs(hl_color_attrs) do
+    attr = i > 0 and ',' .. attr or attr
+    attrs = attrs .. attr
+    i = i + 1
+  end
```
akinsho, line 78: I think this can be done much more simply as `table.concat(vim.tbl_keys(attr_map), ", ")` I think you can even inline it.

lua/bufferline/colors.lua
antoineco: We _are_ completely overriding `DiagnosticError` with a link. However, since the linked highlight group has no `fg` color. Neovim's API return a default `fg` (incorrect) instead of no `fg` at all (correct) unless `link = false` is set.
I say _incorrect_ because the UI doesn't render Neovim's default `fg` color, it ignores `fg` as one would expect, so I'm surprised that the API even returns this default `fg` value (could be a bug).

I think the absence of `fg` on the `DiagnosticError` highlight group in the colorscheme is the reason for both the things you mentioned.

lua/bufferline/colors.lua
antoineco: I totally forgot about Neovim's table helpers. It works indeed, thanks!

```
Failed to run `config` for bufferline.nvim
.../pack/lazy/opt/bufferline.nvim/lua/bufferline/colors.lua:78: unsupported attribute spp for DiagnosticHint, should be
one of bg, sp, fg
```

Done in [7836b304..dff17b60](https://github.com/akinsho/bufferline.nvim/compare/7836b304bfc6eb26d26d9714a6fa42af64d2734e..dff17b6083bbbe9ac5d55c7b84e6b4856be6014b)

lua/bufferline/colors.lua
antoineco: I just looked at my first comment again, and to me the default `link = true` is plain wrong.

I could understand the injection of a default `fg`, as long as the other attributes from the linked group are preserved. But it's much worse that that, with `link = true` _all_ attributes from the linked group are dropped, which is the complete opposite of what the Neovim UI renders on screen.

akinsho: I see 👍🏾 the behaviour was actually purposely added (I was watching the PR that added it) because this is how vim ":highlight" behaves it does this inheritance/melding of highlights so is the "expected" behaviour but understandably not what is desired here. 

I think this is likely going to result in a slew of issues about "weird" highlighting in short order but I'm willing to try it and see what happens. One last thing I think before merging is if you could add the changes here as 2 separate commits one that does the `link = false` and the rest of the changes so I can add both commits and revert the like one if necessary (hopefully it won't be)

antoineco: I did an experiment after running `hi! clear`:

image

image

```vim
:hi! link PreProc Error
```

image

```console
# :=vim.api.nvim_get_hl(0, { name = 'PreProc' })
{
  ctermfg = 81,
  fg = 16744703,
  link = "Error"
}
```

```console
# :=vim.api.nvim_get_hl(0, { name = 'PreProc', link = false })
{
  bg = 16711680,
  ctermbg = 9,
  ctermfg = 15,
  fg = 16777215
}
```

So the issue isn't limited to unset attributes such as `fg` vs. `sp`, etc.
The Lua API _consistently_ returns the wrong attributes unless `link = false` is passed.

antoineco: @akinsho PR split into two commits as requested 👍

akinsho (approved):
LGTM 👍🏾 thanks for your work on this

PR #739, opened 2023-04-24
feat(ui): add `padded_slope` style

ansidev:
## Proposed change

- [x] Add new separator style `padded_slope`.


ansidev:
> @ansidev what terminal is this? I'm wondering if there are specific terminal settings that could increase the spacing without the extra code but it's not a big deal to add I just wish it wasn't necessary to keep adding "padded" variants of things

@akinsho I'm using Alacritty 0.12.0

PR #742, opened 2023-04-25
adding indicator_visible to highlight docs

casonadams:


doc/bufferline.txt
```diff
@@ -810,6 +810,16 @@ NOTE: you can specify colors the same way you specify colors for
...
             },
+            tab_separator = {
+              fg = separator_background_color,
+              bg = background_color,
+            },
+            tab_separator_selected = {
+              fg = separator_background_color,
+              bg = visible_bg,
+              sp = underline_sp,
+              underline = has_underline_indicator,
```
akinsho, line 821: @casonadams these should have a value similar to other colour groups rather than having the lua variables the actual docs

akinsho (changes requested):
@casonadams thanks for the PR, just need to remove the lua variables from the docs and just use the same generic value as the other examples for `underline` that should be a boolean though not a colour value.

akinsho:
Hi thanks for the PR I've made the tweaks I suggested and pushed up a commit. Thanks for bringing this to light though

PR #745, opened 2023-05-01
docs(help): update buffer_close_icon

zaxutic:
In 83b4d0b, the default `buffer_close_icon` was updated to reflect the breaking changes in Nerd Fonts, however the documentation was not updated to use this new default in its example config. This PR updates that to make them consistent.


akinsho:
Thanks 👍🏾

PR #750, opened 2023-05-08
feat(duplicates): only treat as duplicate if in the same group

rickyelopez:


rickyelopez:
@akinsho could it be this simple? Do any tests need to be updated?

akinsho (changes requested):
@rickyelopez looks good. Haven't tried it but looks like it might be that simple. Would be good to have a test case which tests just this function by passing it two duplicated buffers with the same group and verifying that it works. The plugin uses plenary's testing library. The CI is an example of how that runs.


Also I think it's pretty important that this is gated behind an option. I'm personally happy to have the addition but would want it off since if I had two files like `init.lua` tbh regardless of whether or not they are in different groups I'd definitely want to see some sign of how they are different.

rickyelopez:
@akinsho any thoughts here?

akinsho:
@rickyelopez implementation looks good. I just haven't had a lot of time to spend on open source recently. I was thinking also about if I could find a more concise name for the option as it's quite a mouthful at the moment but didn't really find anything

rickyelopez:
No worries! Doesn't seem like anybody else has requested this, and it's working for me this way, so I'm find to leave it in a PR until you get back to this project

> find a more concise name for the option as it's quite a mouthful at the moment but didn't really find anything

Heh, same. Happy to change it if you think of anything, but I came up empty

akinsho (approved):
LGTM

PR #757, opened 2023-05-20
fix(ui): Exclude hidden buffers from the list

mskelton:
When using the `always_show_bufferline` in conjunction with `null-ls` for formatting, the plugin will break the cursor location when formatting. This happens due to the fact that null-ls creates a temporary scratch buffer, runs the formatter, then destroys the temp buffer. In the middle of this, bufferline will see multiple buffers and for a moment attempt to render the bufferline, then hide it again. This seems to be breaking cursor position.

The fix I applied to this was to exclude hidden buffers from the total buffer count. I'm curious your thoughts on the implementation and if there would be any issues from this change.

### DIFF
--- a/lua/bufferline/utils/init.lua
+++ b/lua/bufferline/utils/init.lua
@@ -132,7 +132,14 @@ function M.is_valid(buf_num)
 end
 
 ---@return integer
-function M.get_buf_count() return #fn.getbufinfo({ buflisted = 1 }) end
+function M.get_buf_count()
+  local buffers = vim.tbl_filter(
+    function(buf) return vim.fn.getbufvar(buf.bufnr, "&bufhidden") ~= "hide" end,
+    fn.getbufinfo({ buflisted = 1 })
+  )
+
+  return #buffers
+end
 
 ---@return integer[]
 function M.get_valid_buffers() return vim.tbl_filter(M.is_valid, vim.api.nvim_list_bufs()) end


akinsho:
Hi @mskelton,

So I'm not sure what you mean by "breaking the cursor". I genuinely think there is literally no logic in this plugin that deals with the cursor position. If rendering a buffer in the tabline is causing changes to the cursor this seems like a bug in neovim

mskelton:
Here is a screen recording showing the behavior. The first half is the broken behavior, the second half is the fixed behavior. The issues happens due `showtabline` briefly changing values then quickly changing back as a result of the extra temporary buffer.

https://github.com/akinsho/bufferline.nvim/assets/25914066/7731580b-c6eb-4754-88e8-61141fa06572

akinsho:
Thanks for the video @mskelton, I still think the root cause here is neovim because I really don't understand how even quickly switching the value of `showtabline` would cause the cursor to move, this seems to me like a bug. Those things should be unrelated so the fact that one influences the other seems to be an issue.

mskelton:
@akinsho It seems that bufferline is in part the culprit as loading neovim with any plugins or config and running `:set showtabline=2` doesn't change the cursor position.

https://github.com/akinsho/bufferline.nvim/assets/25914066/b3ce6061-4bf1-4600-9212-9677146eab54

PR #758, opened 2023-05-20
feat(commands/go_to): go to the last element if index out of bounds

utkarshgupta137:
This is how tabs work in most apps (at least on macOS).
I can put this behind a config option if required.

### DIFF
--- a/lua/bufferline/commands.lua
+++ b/lua/bufferline/commands.lua
@@ -130,8 +130,8 @@ function M.go_to(num, absolute)
   num = type(num) == "string" and tonumber(num) or num
   local list = absolute and state.components or state.visible_components
   local element = list[num]
-  if num == -1 then element = list[#list] end
-  if element then open_element(element.id) end
+  if num == -1 or not element then element = list[#list] end
+  open_element(element.id)
 end
 
 ---@param current_state bufferline.State


akinsho (commented):
This generally seems like a good change although I'm thinking a little about the case where someone does something completely wrong maybe due to a bug in their config imagine they only have 3 buffers open and they try to `go_to(22)` this is wildly incorrect. It's strange that it would move at all. Going to the last one in this case isn't really a compromise to my mind since it ostensibly still super far off what the user intended.

I think in this case it's better to just do nothing since the end isn't a good approximation in this case. 🤔 if they just went over then it might make sense, although how far over etc. isn't really something to factor in 🤔

utkarshgupta137:
Yes, I realized while creating this patch that there are a few differences compared to browsers. For one, there is no way to say go to tab 22. Secondly, Go to 9 is actually go to the last tab.
That said, I don't think anyone would actually use `go_to(22)` because you can only fit so many buffers on the screen at once. For me, I've set `tab_size = 0` & I can rarely get more than 10 buffers at once (with Neo-tree open).

akinsho (commented):
Not a 100% sure about the behaviour of going to end the end if someone does something wild. I still feel like that should just error but I guess this can be reverted/hopefully there's no harm in it re. intuitiveness

PR #759, opened 2023-05-20
feat(commands): add option to wrap when moving buffers at ends

### LINKED ISSUE #756: [Bug]: BufferLineMovePrev/Next doesn't work when it is the first/last buffer

### What happened?

Title

If I'm on the first buffer, then BufferLineMovePrev should move the buffer to the last position.
If I'm on the last buffer then BufferLineMoveNext should move the buffer to the first position.

### Config

lazy:
```lua
{
  "akinsho/bufferline.nvim",
  event = "VeryLazy",
  keys = {
    { "<leader>bp", "<Cmd>BufferLineTogglePin<CR>", desc = "Toggle pin" },
    { "<leader>bP", "<Cmd>BufferLineGroupClose ungrouped<CR>", desc = "Delete non-pinned buffers" },
    { "<leader>bh", "<cmd>BufferLineMovePrev<cr>", desc = "Move Left" },
    { "<leader>bl", "<cmd>BufferLineMoveNext<cr>", desc = "Move Right" },
  },
  opts = {
    options = {
      close_command = function(n) require("mini.bufremove").delete(n, false) end,
      right_mouse_command = function(n) require("mini.bufremove").delete(n, false) end,
      diagnostics = "nvim_lsp",
      show_buffer_close_icons = false,
      separator_style = "slope",
      diagnostics_indicator = function(_, _, diag)
        local icons = require("lazyvim.config").icons.diagnostics
        local ret = (diag.error and icons.Error .. diag.error .. " " or "")
          .. (diag.warning and icons.Warn .. diag.warning or "")
        return vim.trim(ret)
      end,
      offsets = {
        {
          filetype = "neo-tree",
          text = "Neo-tree",
          highlight = "Directory",
          text_align = "left",
        },
      },
    },
  },
}
```

### Additional Information

Using latest LazyVim

### commit

_No response_

utkarshgupta137:
Fixes #756.

### DIFF
--- a/lua/bufferline/commands.lua
+++ b/lua/bufferline/commands.lua
@@ -147,6 +147,16 @@ function M.get_current_element_index(current_state, opts)
   end
 end
 
+---@param current_state bufferline.State
+---@return number
+local get_last_pinned_index = function(current_state)
+  for index, item in ipairs(current_state.components) do
+    local element = item:as_element()
+    if not groups.is_pinned(element) then return index - 1 end
+  end
+  return 0
+end
+
 --- Move the buffer at index `from_index` (or current index if not specified) to position `to_index`
 --- @param to_index number negative indices are accepted (counting from the right instead of the left, e.g. -1 for the last position, -2 for the second-last, etc.)
 --- @param from_index number?
@@ -169,8 +179,25 @@ end
 
 --- @param direction number
 function M.move(direction)
-  local index = M.get_current_element_index(state)
-  M.move_to(index + direction, index)
+  local index, element = M.get_current_element_index(state)
+  local next_index = index + direction
+  if not config.options.move_wraps_at_ends or not index then return M.move_to(next_index, index) end
+
+  local last_pinned_index = get_last_pinned_index(state)
+  if groups.is_pinned(element) then
+    if next_index <= 0 then
+      next_index = last_pinned_index
+    elseif next_index > last_pinned_index then
+      next_index = 1
+    end
+  else
+    if next_index <= last_pinned_index then
+      next_index = #state.components
+    elseif next_index > #state.components then
+      next_index = last_pinned_index + 1
+    end
+  end
+  M.move_to(next_index, index)
 end
 
 --- @param direction number
--- a/lua/bufferline/groups.lua
+++ b/lua/bufferline/groups.lua
@@ -309,7 +309,7 @@ local group_by_name = group_by("name")
 local group_by_priority = group_by("priority")
 
 ---@param element bufferline.TabElement
-local function is_pinned(element) return get_manual_group(element) == PINNED_ID end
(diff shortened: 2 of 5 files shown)


lua/bufferline/groups.lua
```diff
@@ -309,7 +309,7 @@ local group_by_name = group_by("name")
 local group_by_priority = group_by("priority")
 
 ---@param element bufferline.TabElement
-local function is_pinned(element) return get_manual_group(element) == PINNED_ID end
+function M.is_pinned(element) return get_manual_group(element) == PINNED_ID end
```
akinsho, line 312: I actually recently made this function private to this module and that caused some kerfuffle as I didn't want it to be used externally by users. I'm wondering why it is required for what you are doing?

lua/bufferline/commands.lua
```diff
@@ -147,6 +147,16 @@ function M.get_current_element_index(current_state, opts)
   end
 end
 
+---@param current_state bufferline.State
+---@return number
+local get_last_pinned_index = function(current_state)
```
akinsho, line 152: I'm curious why this function is required at all, I had assumed the change would just involve setting the next index to the start or finish rather than interacting with groups or pinned elements.

akinsho (commented):
@utkarshgupta137 just got round to reviewing this. I have questions about why the pinned functionality from `groups` has to be re-exposed and why there is so much logic involving pinned buffers. I kind of expected this to not involve any of that functionality

utkarshgupta137:
> @utkarshgupta137 just got round to reviewing this. I have questions about why the pinned functionality from `groups` has to be re-exposed and why there is so much logic involving pinned buffers. I kind of expected this to not involve any of that functionality

Assume I've got 7 buffers open & the first 3 are pinned. So 4th is the first unpinned buffer.
If I try to move the 1st buffer left, it should move to position 3.
If I try to move the 3rd buffer left, it should move to position 1.
If I try to move the 4th buffer left, it should move to position 7.
If I try to move the 7th buffer right, it should move to position 4.

If it instead moved the 7th buffer to position 1, then pinning breaks in various ways. Same if I move the 3rd buffer to the 4th position. I think the code (rightly) assumes in some places that all the pinned buffers would take the first few slots before any unpinned buffers.

lua/bufferline/groups.lua
akinsho, line 312: @utkarshgupta137 can you prefix this with an underscore e.g. `M._is_pinned` so that it's clear that this is private. I don't want users to start using it again.

akinsho (commented):
Okay tried it out locally and it seems to work fine and respect the pinned positions and groups 👍🏾 just one small comment about making the `is_pinned` function start with an `_` so at least it's a signal that it shouldn't be used by users but I'm sure someone will anyway

akinsho:
Merged, thanks for the PR. I've updated the is pinned method to be private as well

PR #760, opened 2023-05-21
refactor(utils): use helper variable consistently

tobealive:
Makes consistent use of the strwidth helper var declared at the top of the file.
Simple refactor. No real work done here.

Great work with the plugin, I like the performance especially. Thanks a lot for providing value @akinsho.


akinsho (approved):
LGTM thanks 👍🏾

Issue #317, opened 2022-01-26
[Question]: How to close other buffers ?


holicc:
want to close other buffers except current buffer

akinsho:
@holicc please have a look at `:help :only` in vim as there is already an existing vim command for this. Gonna close this out now since that's the answer and isn't something specific to this plugin

pirey:
`:only` only close other windows, and it does not close other buffers.

we can take advantage of close left and close right command for workaround

```lua
      vim.keymap.set(
        "n",
        "<leader>bo",
        "<cmd>BufferLineCloseLeft<cr><cmd>BufferLineCloseRight<cr>",
        { desc = "Close other buffers" }
      )

```

PR #774, opened 2023-06-17
feat(command): add BufferLineCloseOthers command

pirey:
This PR addresses the issue https://github.com/akinsho/bufferline.nvim/issues/317

In the issue, it was suggested to use `:only`, but `:only` will only close other windows, not buffers.

This new command can be used to close all buffers except the current one.


lua/bufferline/commands.lua
```diff
@@ -253,6 +253,19 @@ function M.close_in_direction(direction)
   ui.refresh()
 end
 
+--Close other buffers
+function M.close_others()
+  local index = M.get_current_element_index(state)
+  if not index then return end
+
+  for i, item in ipairs(vim.list_slice(state.components)) do
```
akinsho, line 261: @pirey why does this do a slice first? rather than just iterate the components?

pirey: my bad, i was just trying things out.
i have updated it @akinsho

lua/bufferline/commands.lua
```diff
@@ -253,6 +253,19 @@ function M.close_in_direction(direction)
   ui.refresh()
 end
 
+--Close other buffers
+function M.close_others()
+  local index = M.get_current_element_index(state)
+  if not index then return end
+
+  for i, item in ipairs(state.components) do
+    if not (i == index) then
```
akinsho, line 262:
```suggestion
    if i ~= index then
```

akinsho (commented):
@pirey the last thing this needs is to be added to the help documentation in `bufferline.txt` so other users know this exists

pirey:
@akinsho 
I have updated the doc file

akinsho (approved):
LGTM now thanks 👍🏾

PR #778, opened 2023-06-25
docs: fix commas in highlights examples

MisterFISHUP:
Add commas in some examples so that users can copy them directly without having errors like

```lua
fg = '<colour-value-here>',
bg = '<colour-value-here>',
sp = '<colour-value-here>'  -- missing a comma here
bold = true,
italic = true,
```


akinsho (approved):
LGTM 👍🏾

PR #780, opened 2023-06-27
fix: store paths in g:BufferlinePositions

AThePeanut4:
This fixes an issue where restoring the custom sort order from a saved session doesn't work correctly most of the time, because it stores buffer ids in `g:BufferlinePositions` which are not preserved by `:mksession`. The fix is to store the full paths of the buffers instead. `vim.json` is required to stringify the list as `:mksession` only stores string/number values.

Split out from #777.

### DIFF
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -116,7 +116,7 @@ local function setup_autocommands(conf)
     api.nvim_create_autocmd("SessionLoadPost", {
       pattern = "*",
       group = BUFFERLINE_GROUP,
-      callback = function() state.restore_positions() end,
+      callback = function() state.custom_sort = utils.restore_positions() end,
     })
   end
   if not options.always_show_bufferline then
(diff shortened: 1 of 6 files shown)


akinsho (commented):
@AThePeanut4 this change looks good to me generally. Haven't had time to test it myself but would be great to have a test to just confirm that this logic works.

It could be something as simple as setup bufferline in a test (see `bufferline_spec.lua` then open 3 files e.g. `vim.cmd.edit('example-1.txt')` etc. Then run `save_positions` then run `restore_positions` and check that the state is what you expect

AThePeanut4:
Alright, I'll see what I can do.

AThePeanut4:
Alright I moved the functions to utils and added a test to `utils_spec.lua`. This PR doesn't really change anything outside of `save|restore_positions` so that's all I've tested.

akinsho:
Merged, thanks for taking this on @AThePeanut4

PR #781, opened 2023-07-04
feat(ui): trunc marker highlights

mluna-again:
I added a new highlight for the trunc marker, because it was using the `BufferLineFill` hl.
I also updated the docs.


akinsho (approved):
LGTM 👍🏾

Issue #291, opened 2022-01-06
[Feature request]:  Is there a way to numbering with N.th?


66RING:
### What happened?

Thanks for your awesome work.👍

I want bufferline always be numbered 1th,2th,3th...Nth, so that I can easily get a number and use BufferLineGoToBuffer N.

always numbering with N.th:

```
<- (30) | 1.buf1 | 2.buf2 | 3.buf3 |(24) ->

| 1.buf2 | 2.buf1 | 3.buf3 |(24) ->

<- (30) | 1.buf3 | 2.buf1 | 3.buf2 |
```

### What did you expect to happen?

```
| 1.buf1 | 2.buf2 | 3.buf3 | 
```

1. go to buf2
2. number is 2
2. : BufferLineGoToBuffer 2

```
| 1.buf2 | 2.buf1 | 3.buf3 |
````

1. go to buf1
2. number is 2
2. : BufferLineGoToBuffer 2

**Comment by akinsho** [comment]
Hi @66RING,

I'm not sure I understand, currently the numbers are up to the user and can be the visible number or the buffer number, if you use the buffer number you can just use `:b<bufnum>` otherwise `:BufferLineGotoBuffer <visiblenum>` will take you to the visible buffer. I can't tell from the above description what doesn't work or which way things should.

**Comment by 66RING** [comment]
My bad. My english is not good enough and I missed the point. I mean using the `numbers=visiblenum` may get a better workflow. The demand should be the same as this #249.

kind of

```lua
require('bufferline').setup{
    options.numbers = visiblenum
}
```

My shortcut:

```lua
map <BS> :lua require"bufferline".go_to_buffer(vim.v.count)<CR>
```

If use `:b<bufnum>` , I need to press key more than twice often. Such as `22<BS>`, 22 is the buffer number and I need to press three key `2, 2, <BS>`.

![image](https://user-images.githubusercontent.com/49358576/148379310-43f0d90a-2a21-4f35-974a-fe0137543fa3.png)


If use `BufferLineGoToBuffer`, since `BufferLineGoToBuffer 4` will go to the **4th** buffer, I have to figure out which position is the 4th.

So, if  `options.numbers = visiblenum` I will know where is the 4th position at a glance and press two key only `{1..7}<BS>`.

carsin:
any update on this? i also find the default behavior very unnatural

akinsho:
@carsin or @66RING either of you are very welcome to submit a PR to help implement this. 
[...]

Issue #249, opened 2021-11-03
Are ordinal numbers static?


rod-stuchi:
### What happened?

![image](https://user-images.githubusercontent.com/19471811/140195648-1824d322-9142-4afe-8542-f897efbad953.png)

All good up to here

![image](https://user-images.githubusercontent.com/19471811/140195793-cedb9ef2-dd26-44db-ad82-9d14bcdc31ca.png)

But, if I hit `<Leader7>`  it stays at file006 `ordinal 6`

if I hit `<Leader1>`  it goes to file004, with `ordinal 4`

![image](https://user-images.githubusercontent.com/19471811/140196216-7d76220d-fc45-453e-b993-28e9735adbea.png)

### What did you expect to happen?

Should this ordinals be dynamic?

### Additional Information

...

akinsho:
@rod-stuchi no they aren't static, they represent the visible position of the buffer in the bar so as you move through visible buffers and the list shifts the numbers will shift

rod-stuchi:
Thanks @akinsho 
So in this case, it could be a bug in mappings? 
In the flow up above, if I hit `<Leader>1` it goes to file004 (ordinal 4), but the correct was to select file001 (ordinal 1).

IMO, I think it is cooler to select ordinal according to what is visible, if file001 was offset, file004 should be (ordinal 1) but the ordinal number should be updated visually, but the ordinal number get updated only by moving its position.
I don't know if this is even possible.

rod-stuchi:
Hi @akinsho , this video (with screenkey) shows better what happen.

https://user-images.githubusercontent.com/19471811/140356222-8cc3f273-f600-4ebb-a2a4-9a12ddbf4077.mp4

My mappings, `Leader = ','`
```
keymap('n', '<Leader>1', '<Cmd>BufferLineGoToBuffer 1<CR>', opts)
keymap('n', '<Leader>2', '<Cmd>BufferLineGoToBuffer 2<CR>', opts)
keymap('n', '<Leader>3', '<Cmd>BufferLineGoToBuffer 3<CR>', opts)
keymap('n', '<Leader>4', '<Cmd>BufferLineGoToBuffer 4<CR>', opts)
keymap('n', '<Leader>5', '<Cmd>BufferLineGoToBuffer 5<CR>', opts)
keymap('n', '<Leader>6', '<Cmd>BufferLineGoToBuffer 6<CR>', opts)
[...]

PR #797, opened 2023-07-31
Add relative buffer number for currently visible buffer

rameshsanth:
Here is a possible fix for #249 , #291 . This is more like a proof of concept of how this can be achieved.

The fix involves add bufferline index to the buffer name as <bufid>.<buflineIndex>.
In the screenshots below, you can see the filenames start with <X>.1 onwards 

It works in my setup with my configuration. However I'm not sure how this would work with various options that bufferline provides.

I welcome comments.

Screenshot 2023-07-30 at 7 33 49 PM

Upon selecting BufferLineGoToBuffer 1
Screenshot 2023-07-30 at 7 35 34 PM

### DIFF
--- a/lua/bufferline/ui.lua
+++ b/lua/bufferline/ui.lua
@@ -538,19 +538,25 @@ end
 --- and converts them into an nvim tabline format string i.e. `%#HL#text`. It handles cases
 --- like applying global or local attributes like click handlers. As well as extending highlights
 ---@param component bufferline.Segment[]
-local function to_tabline_str(component)
+local function to_tabline_str(component, tabline_index)
   component = component or {}
+  local highlight_found = false
   local str = {}
   local globals = {}
   extend_highlight(component)
   for _, part in ipairs(component) do
     local attr = part.attr
     if attr and attr.global then table.insert(globals, { attr.prefix or "", attr.suffix or "" }) end
+    local text = part.text or ""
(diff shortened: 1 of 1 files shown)


otavioschwanck:
@rameshsanth @akinsho any prevision for when it will be merged?  The ordinal here don't have the correct order.

image

would be awesome to always keep the order (helps on navigation)

akinsho:
This one was raised more as a proof of concept and not definitive IMO. I haven't had much time to dig into these sorts of things in the last few weeks. I'm really focusing on my side projects, not on open source stuff at the moment. More than any other feature, the numbers feature creates a lot of work for me and frankly I'm fairly convinced I should deprecate it given how many issues I get about numbers.

Looking at it I don't think this is an easy merge. It seems like something to discuss because this will definitely mess up some other people's flow and I'm not entirely sure about what the problem clearly is. I know that they're out of order but I'm not sure what people in general want out of the numbers because it never seems to be enough or right enough for everyone

akinsho:
I get pinged with every update of this PR so I'm going to close it as is as I don't think it's in a mergeable state. I'm open to the idea but not this implementation. It really muddies the waters by directly hardcoding highlight colour names in an area where those kind of specifics shouldn't be. If you or anyone else is willing to try another PR then please have at it but ideally maintain the separation of concerns and avoid overly imperative logic 🙏🏾 sorry but fleshing out how to approach it would be a non zero amount of work and will have to be part of the job of the implementor figuring this out

PR #800, opened 2023-08-02
refactor(offsets): use wininfo to calculate offset

### LINKED ISSUE #789: [Bug]: Offset only uses one window

### What happened?

When multiple offsets are defined, then opening a new window will only apply the offset configuration to one window, usually the leftmost one or the last one opened. It behaves oddly.

For example when having an offset for filetype "help" and an offset for filetype "NvimTree", then opening NvimTree will produce the correct behavior. If you then however open the help with `:vertical help`, the help window gets the offset configuration applied but the nvim-tree gets it's text deleted.

To reproduce:

1. Open nvim and only have one buffer open that doesn't have an offset setting for it's filetype
2. Open nvim-tree
image

3. Open vertical help
image

4. Notice help doesn't get the title and instead the bufferline starts over the help window
5. Close nvim-tree
image

6. Notice that help now gets the offset applied where before it did not

Windows that open on the right, such as tagbar, also behave oddly when opening and closing them mixed with opening and closing multiple windows that open on the left, such as undotree.

Vertically split windows always get their header text and all the offsets are added for the bufferline, except of course for the offsets for windows on the right, which should be handled separately

### Config

```lua
-- ...
offsets = {
    {
        filetype = "NvimTree",
        text = function()
            return "NVIMTREE TEXT"
            -- return vim.fn.getcwd()
        end,
        -- highlight = "Directory",
        text_align = "left",
        separator = true
    },
    {
        filetype = "NERDTree",
        text = function()
            return vim.fn.getcwd()
        end,
        highlight = "Directory",
        text_align = "left",
        separator = true
    },
    {
        filetype = "undotree",
        text = "Undo Tree",
        text_align = "left",
        separator = true
    },
    {
        filetype = "diff",
        text = "Diff",
        text_align = "left",
        separator = true
    },
    {
        filetype = "tagbar",
        text = "Tags",
        text_align = "left",
        separator = true
    },
    {
        filetype = "help",
        text = "Help",
        text_align = "left",
        separator = true
    }
},
-- ...
```

### Additional Information

N/A

### commit

_No response_

akinsho:
fixes #789


nullptr-yxw:
It still cannot work as expect when the offset is the only window

https://github.com/akinsho/bufferline.nvim/assets/96286337/39cfdc3b-c795-4f55-b3e8-8dd2db552f86

my configuration of options.offsets is
```lua
offsets = {
    {
        filetype = "NvimTree",
        text = "File Explorer",
        highlight = "Directory",
        text_align = "left",
        separator = true,
    },
},
```

PR #806, opened 2023-08-25
feat: Support `name_formatter` for unnamed buffers

mskelton:
This adds a new config option `unnamed_title` to support changing the name of unnamed buffers from `[No Name]` to anything the user wants (e.g. `Untitled`).

### DIFF
--- a/lua/bufferline/models.lua
+++ b/lua/bufferline/models.lua
@@ -176,10 +176,12 @@ function Buffer:new(buf)
   if buf.path and #buf.path > 0 then
     name = fn.fnamemodify(buf.path, ":t")
     name = is_directory and name .. "/" or name
-    if buf.name_formatter and type(buf.name_formatter) == "function" then
-      name = buf.name_formatter({ name = name, path = buf.path, bufnr = buf.id }) or name
-    end
   end
+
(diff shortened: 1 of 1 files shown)


akinsho:
@mskelton I think you can accomplish this busy using the `name_formatter` field you can just specify a function and if the path of the buffer passed in is an empty string then use whatever name you want

mskelton:
`name_formatter` is not called for unnamed buffers.

```lua
  local name = "[No Name]"
  if buf.path and #buf.path > 0 then
    name = fn.fnamemodify(buf.path, ":t")
    name = is_directory and name .. "/" or name
    if buf.name_formatter and type(buf.name_formatter) == "function" then
      name = buf.name_formatter({ name = name, path = buf.path, bufnr = buf.id }) or name
    end
  end

```

akinsho:
@mskelton I think the solution then IMO would be adapting the logic to work with no name buffers as well so there's one versatile solution rather than two

mskelton:
@akinsho I changed to allow `name_formatter` to work with unnamed buffers.

akinsho (approved):
LGTM 👍🏾 thanks for making the changes

PR #812, opened 2023-09-07
fix: remove `missing required fields` diagnostic from config

tj-moody:
As discussed in the mentioned issue, it appears that some changes to lua_ls mean that not including non-optional fields in a config table throws a diagnostic warning, as shown in the screenshot. As the issue mentions, it's possible to disable this behavior globally or with a `---@diagnostic disable-next-line missing-fields` comment before the table, but I think it would still be greatly desirable to have bufferline not throw warnings for non-mandatory fields by default.

I've tried my best to only make the non-mandatory fields optional, but I probably over or under-modified the types, so please feel free to add/remove `?`s around the types file.

https://github.com/LuaLS/lua-language-server/issues/2214#issuecomment-1686426027
Screenshot 2023-09-07 at 11 48 57 AM

### DIFF
--- a/lua/bufferline/types.lua
+++ b/lua/bufferline/types.lua
@@ -4,15 +4,15 @@
 ---@field logging boolean
 
 ---@class bufferline.GroupOptions
----@field toggle_hidden_on_enter boolean re-open hidden groups on bufenter
+---@field toggle_hidden_on_enter? boolean re-open hidden groups on bufenter
 
 ---@class bufferline.GroupOpts
----@field options bufferline.GroupOptions
+---@field options? bufferline.GroupOptions
 ---@field items bufferline.Group[]
 
 ---@class bufferline.Indicator
----@field style "underline" | "icon" | "none"
----@field icon string?
+---@field style? "underline" | "icon" | "none"
+---@field icon? string?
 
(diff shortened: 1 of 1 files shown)


tj-moody:
As a note here is an analogous pr in for nvim-cmp, which the original issue was about: https://github.com/hrsh7th/nvim-cmp/pull/1694

lua/bufferline/types.lua
```diff
@@ -4,15 +4,15 @@
...
 ---@class bufferline.GroupOpts
----@field options bufferline.GroupOptions
+---@field options? bufferline.GroupOptions
 ---@field items bufferline.Group[]
 
 ---@class bufferline.Indicator
----@field style "underline" | "icon" | "none"
----@field icon string?
+---@field style? "underline" | "icon" | "none"
+---@field icon? string?
```
claudiofreitas, line 15: If adding `?` after the field/param name, then maybe it might be a good idea to also remove the `?` after the type to make it consistent.

```suggestion
---@field icon? string
```

I think there are also others like this in this file:
lines 15, 37, 42, 50`*`, 181`*`

`*` = for those lines I am not completely sure

akinsho (approved):

PR #814, opened 2023-09-10
config: Added option to override the base colors

ofirgall:
Hey, I'm not sure if its a desired configuration option but I added the option to override the base colors of the calculation rather override the end result.

It helped me get better color for my bufferline, if it's a desired changed I can add documentation to the help file.

### DIFF
--- a/lua/bufferline/config.lua
+++ b/lua/bufferline/config.lua
@@ -151,7 +151,7 @@ function Config:is_tabline() return self:mode() == "tabs" end
 ---Derive the colors for the bufferline
 ---@param preset bufferline.StylePreset | bufferline.StylePreset[]
 ---@return bufferline.Highlights
-local function derive_colors(preset)
+local function derive_colors(preset, base_colors)
   local hex = colors.get_color
   local tint = colors.shade_color
   if type(preset) ~= "table" then preset = { preset } end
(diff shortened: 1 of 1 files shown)


akinsho:
@ofirgall thanks for the PR it's not a bad idea but I feel the complexity around the colouring of bufferline has exploded since I first created it as it's simply impossible to style anything in neovim in a way that will appeal to everyone. This isn't an unreasonable change but I just think the barrier to adding any new highlighting logic has to be super high as it's already become fairly painful to maintain as every change breaks something for someone somewhere depending on some very specific nuance of the current highlighting.

ofirgall:
I agree that you can't create a logic that would appeal to everyone, that's the reason I did this change. It allows the user to override the variables of the highlight logic in order to provide more control over the outcome, but it didn't change the underlying logic.

I know the idea isn't future-proof because it relies on the internal highlighting logic. In my mind I figured this section of the code isn't changed much because of the reasons you mentioned. I would like to hear your take on the idea, and if you would implement it differently?

akinsho:
@ofirgall the implementation is not an issue, I think for me with my plugins and neovim in general it's hard to avoid endless configuration options and endless variations on everything to account for x,y,z etc. the list literally never ends as my time maintaining this plugin has taught me. After you introduce this change, despite the fact that every single group can be changed and (with this change) the base palette someone else will still want something else. So I think from a maintenance perspective part of my responsibility is to draw a line and say this is enough configuration.

In this case the colours that make up the base palette are intentionally opinionated, tbh deciding on any specific colours in nvim is always very opinionated but I had a very specific appearance in mind and the escape hatch of choosing every highlight group yourself I think is a sufficient way out.

PR #816, opened 2023-09-15
Ignore disabled diagnostics

sidequestboy:


akinsho (changes requested):
@sidequestboy this looks good to me but the method you are using `is_disabled` was only added in nvim `v0.9.0` so if I merge this anyone on an older version of neovim will have issues. Can you add a check here to ensure that that method exists before using it, if not don't use it.

lua/bufferline/diagnostics.lua
```diff
@@ -70,8 +70,10 @@ local get_diagnostics = {
   nvim_lsp = function()
     local results = {}
     local diagnostics = vim.diagnostic.get()
+    -- is_disabled introduced in 0.9
+    local is_new_nvim = fn.has("nvim-0.9") > 0
```
akinsho, line 74: @sidequestboy I think this can be simplified and made more robust by just checking that the function exists. This is what I do elsewhere e.g.
if `vim.diagnostic.is_disabled ~= nil` rather than checking for the specific version. Also would be good to have a TODO comment for the future so I can come in and remove this eventually 🙏🏾

sidequestboy:
thanks for catching that, lmk if this looks good

akinsho (approved):
LGTM 👍🏾 thanks for tweaking this

Issue #191, opened 2021-08-20
[Bug]: Bufferline don't show overidden icons which are set by nvim-web-devicons


CaptainKO:
### What happened?

 default = true,
}

### What did you expect to happen?

```lua
-- Test config
print(vim.inspect(devicons.get_icon('test.component.spec.ts'))) -- print: 
```

### Additional Information

[nvim-web-devicons config](https://github.com/kyazdani42/nvim-tree.lua/pull/423#issue-658556199)

akinsho:
Hi @CaptainKO, can you confirm when you are setting this override i.e. are you sure it is being set before bufferline starts?

CaptainKO:
I put it int the config callback.
I as can see right now `nvim-tree` and `coc-explorer` work fine.

```lua

  use {
    "kyazdani42/nvim-web-devicons",
    config = function()
      local override = {
        ["spec.ts"] = {
          icon = "",
          color = "#ffc600",
          name = "TsTest",
        },
      }

      require("nvim-web-devicons").setup {
        -- your personnal icons can go here (to override)
        -- DevIcon will be appended to `name`
        -- override = vim.tbl_extend("keep",override ,devicons.get_icons()),
        override = override,
        -- globally enable default icons (default to false)
        -- will get overriden by `get_icons` option
        default = true,
      }
    end,
  }
```

akinsho:
@CaptainKO sorry I should have clarified, does bufferline load *before* `nvim-web-devicons` or not? i.e. if you are applying the override after bufferline has already loaded, then that might be why you don't see them applied

akinsho:
I've also made a few changes to how icons are required. Can you test it out and see if it works now

CaptainKO:
Thank you. I will check it out ASAP.

CaptainKO:
I have updated to the latest version and made sure that the `nvim-web-devicons` was configured before loading `bufferline`.
```lua
-- ...
  use {
    "akinsho/bufferline.nvim",
    requires = "kyazdani42/nvim-web-devicons",
    config = config "bufferline",
    after = { "nvim-web-devicons" },
  }
-- ...
```

CaptainKO:
Actually `bufferline` properly shows the overridden icon if the icon was specified only by extension.

```lua
  local override = {
    ["spec.ts"] = { -- not  working
      icon = "",
      color = "#ffc600",
      name = "TsTest",
    },
    ["snippets"] = { -- works fine
      icon = " ",
[...]

PR #817, opened 2023-09-18
fix: display overriden devicons

mscianek:
pretty much a fix for #191

normally  `bufferline` retrieves the file extension using `vim.fn.fnamemodify` with the `:e` option, but that doesn't account for extensions with multiple dots, by passing nil instead of that extension, the inner recursive mechanism of `nvim-web-devicons` is utilized to get the longest matching extension from the filename

before:
before

after:
after


akinsho (approved):
LGTM 👍🏾

akinsho (approved):
LGTM 👍🏾

PR #818, opened 2023-09-20
fix: UI error when UI's `diagnostic.text` is `nil`

amv-dev:
I can't create reproducible example, but this problem periodically happens to me personally. 

After some action (deleting line most frequent) raises an error after which tabs line becomes as neovim default and this plugin needs to be re-initilized.

It also happens if I use `lazy.nvim` as plugin manager and running `Sync` command.

### DIFF
--- a/lua/bufferline/ui.lua
+++ b/lua/bufferline/ui.lua
@@ -497,7 +497,7 @@ function M.element(current_state, element)
     set_id(name, components.id.name),
     spacing({ when = name, highlight = curr_hl.buffer }),
     set_id(diagnostic, components.id.diagnostics),
-    spacing({ when = diagnostic and #diagnostic.text > 0 }),
+    spacing({ when = diagnostic and diagnostic.text and #diagnostic.text > 0 }),
     right_space,
     suffix,
     spacing({ when = suffix }),


amv-dev:
It there something wrong with my PR? It seem to be pretty simple fix, which can't break anything.

akinsho:
> I can't create reproducible example,

I think this is fairly concerning @amv-dev merging a fix for a bug that I cannot reproduce, that no one else has ever reported is not ideal. If I do this as a rule I end up in a position where I'm maintaining a bunch of conditional logic which I genuinely don't know why this is even there.

This PR looks like it will likely mask an issue that should be better understood. If you are able to find the root cause at least or give some clearer indication of what is going on I'd be happy to merge a solution. But a fix that has no clear logic as to why doesn't sound good to me.

amv-dev:
I could only reproduce bug with `lazy.nvim`, but it still happens randomly while editing.

Here is a config for reproducing:
```lua
-- Default `lazy.nvim` initialization script
local lazypath = vim.fn.stdpath("data") .. "/lazy/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
    vim.fn.system({
        "git",
        "clone",
        "--filter=blob:none",
        "https://github.com/folke/lazy.nvim.git",
        "--branch=stable", -- latest stable release
        lazypath,
    })
end
vim.opt.rtp:prepend(lazypath)

require("lazy").setup({
    {
        'akinsho/bufferline.nvim',
        requires = { 'nvim-tree/nvim-web-devicons' },
        opts = {
            options = {
                mode = "tabs",
                diagnostics = "nvim_lsp",
               
                -- Removing this callback will suppress the bug
                diagnostics_indicator = function(count, level, diagnostics_dict, context) end,
            },
        },
    }
})
```
Run default ubuntu image with this config:
`docker run --rm --volume=$HOME/config.lua:/root/.config/nvim/init.lua -it ubuntu /bin/bash`
Inside container:
1. `apt update && apt install -y wget git`
2. `wget https://github.com/neovim/neovim/releases/download/v0.9.2/nvim-linux64.tar.gz`
3. `tar -xf nvim-linux64.tar.gz --strip-components=1 -C /usr`
4. `nvim` . First error must appear right after `lazy.nvim` finishes downloading.
5. Can be reproduced again by restarting `nvim` and executing `Lazy sync` command.

akinsho:
@amv-dev 
`diagnostics_indicator = function(count, level, diagnostics_dict, context) end`
This isn't a correct config for a diagnostics indicator, this is more user error than anything else? this function shouldn't be returning `nil` is this the real context you're experiencing this in

amv-dev:
Real `diagnostic_indicator` is like that:
```lua
diagnostics_indicator = function(count, level, diagnostics_dict, context)
  if diagnostics_dict["error"] ~= nil then return "⛔" end
  if diagnostics_dict["warning"] ~= nil then return "🚩" end
  if diagnostics_dict["hint"] ~= nil then return "❇ " end
end,
```
So, yes, when diagnostic has some another level (not error, warn or hint), this function produces `nil`. And, yes, if I add `return ""` at the end, problem disappears. Thank you for helping.
But I still think that it's better to be able to return `nil` to perform default diagnostic indicator generation in that case.

PR #821, opened 2023-09-24
Fix potential nil access

dsully:
If one uses this method mapped to keys and accidentally hit a number with a non-existing buffer, `element` can be nil:

```
E5108: Error executing lua: ...re/nvim/lazy/bufferline.nvim/lua/bufferline/commands.lua:122: attempt to index local 'element' (a nil value)
stack traceback:
        ...re/nvim/lazy/bufferline.nvim/lua/bufferline/commands.lua:122: in function 'go_to'
```

### DIFF
--- a/lua/bufferline/commands.lua
+++ b/lua/bufferline/commands.lua
@@ -120,7 +120,7 @@ function M.go_to(num, absolute)
   local element = list[num]
   if num == -1 or not element then element = list[#list] end
 
-  if element ~= nil then
+  if element then
     open_element(element.id)
   end
 end
--- a/lua/bufferline/commands.lua
+++ b/lua/bufferline/commands.lua
@@ -119,7 +119,10 @@ function M.go_to(num, absolute)
   local list = absolute and state.components or state.visible_components
   local element = list[num]
   if num == -1 or not element then element = list[#list] end
-  open_element(element.id)
+
+  if element ~= nil then
+    open_element(element.id)
+  end
 end
 
 ---@param current_state bufferline.State


lua/bufferline/commands.lua
```diff
@@ -119,7 +119,10 @@ function M.go_to(num, absolute)
   local list = absolute and state.components or state.visible_components
   local element = list[num]
   if num == -1 or not element then element = list[#list] end
-  open_element(element.id)
+
+  if element ~= nil then
```
akinsho, line 123:
```suggestion
  if element then
```

lua/bufferline/commands.lua
```diff
@@ -119,7 +119,10 @@ function M.go_to(num, absolute)
   local list = absolute and state.components or state.visible_components
   local element = list[num]
   if num == -1 or not element then element = list[#list] end
```
akinsho, line 121: I'm a bit confused why this check wouldn't be enough already? if element is falsy then it should use the last item in the list?

lua/bufferline/commands.lua
dsully: Thanks - I keep forgetting that Lua indexing isn't 0-based.

lua/bufferline/commands.lua
dsully: If I insert:

```
vim.notify(vim.inspect(element))
vim.notify(vim.inspect(list))
```

in between lines 120 and 121, the output is `nil` and `{}` respectively.

akinsho (approved):
LGTM 👍🏾

PR #833, opened 2023-11-19
fix(bufferline): show when there're multipie tabs

vollowx:
When using `scope.nvim`, one tab may has only one buffer and bufferline wouldn't be shown. This MR change the logic so that the bufferline will be shown if there're more than 1 tab and no matter how many buffers it has in this case.

### DIFF
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -87,8 +87,11 @@ end
 
 --- If the item count has changed and the next tabline status is different then update it
 local function toggle_bufferline()
-  local item_count = config:is_tabline() and utils.get_tab_count() or utils.get_buf_count()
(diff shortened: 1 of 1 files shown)


akinsho:
I don't really understand the rationale for this change or why the change is happening here rather than inside scope.nvim if the behaviour is due to some preference whilst using that plugin

vollowx:
When I use `scope.nvim`, buffers with be attached to tabs, which means one tab may only has one buffer while there're multipie tabs, and I want to show the tabline no matter how many buffers a tab has if there're multipie tabs.

The original logic only considers buffers as multipie tabs always means
multipie buffers.

But anyway, this behavior may be a personal custom, so if you think this isn't good for all users, close if need.

PR #839, opened 2023-12-05
use link if specified in custom areas

### LINKED ISSUE #838: [Bug]: bufferline custom area colors not refreshing

### What happened?

I setup an custom area like so:

```
{
  options = {
    custom_areas = {
      left = function()
        local result = {}
        if vim.o.ft == 'vim' then
          table.insert(result, { text = 'VIM', link = 'BufferLineBufferSelected' })
        else
          table.insert(result, { text = 'Other', link = 'BufferLineBufferVisible'})
        end
        return result
      end
    },
  },
}
```

The texts are refreshing between 'VIM' and 'Other', but the highlights are not changing

the highlights change when switching between a vim file and other types of file

### Config

```lua
{
  options = {
    custom_areas = {
      left = function()
        local result = {}
        if vim.o.ft == 'vim' then
          table.insert(result, { text = 'VIM', link = 'BufferLineBufferSelected' })
        else
          table.insert(result, { text = 'Other', link = 'BufferLineBufferVisible'})
        end
        return result
      end
    },
  },
}
```

### Additional Information

        branch main
        commit 9e8d2f6

### commit

_No response_

jackielii:
fix #838 

If use specified a highlight group via link, use it instead of creating a new group

### DIFF
--- a/lua/bufferline/custom_area.lua
+++ b/lua/bufferline/custom_area.lua
@@ -11,9 +11,12 @@ local fmt = string.format
 ---generate a custom highlight group
 ---@param index integer
 ---@param side string
----@param section {fg: string, bg: string, text: string}
+---@param section {fg: string, bg: string, text: string, link: string}
 ---@param bg string?
 local function create_hl(index, side, section, bg)
+  if section.link then
+    return highlights.hl(section.link)
+  end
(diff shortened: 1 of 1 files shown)


akinsho (approved):
LGTM 👍🏾

Issue #843, opened 2023-12-09
[Bug]: Outdated bufferline-functions Documentation


mcDevnagh:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What happened?

The documentation has an outdated means of executing arbitrary functions on buffers: 
[https://github.com/akinsho/bufferline.nvim/blob/ac788fbc493839c1e76daa8d119934b715fdb90e/doc/bufferline.txt#L551](https://github.com/akinsho/bufferline.nvim/blob/ac788fbc493839c1e76daa8d119934b715fdb90e/doc/bufferline.txt#L551)
```vim
CUSTOM-FUNCTIONS                                     *bufferline-functions*

A user can also execute arbitrary functions against a buffer using the
`buf_exec` function. For example

>lua
    require('bufferline').buf_exec(
        4, -- the forth visible buffer from the left
        user_function -- an arbitrary user function which gets passed the buffer
    )

    -- e.g.
    function _G.bdel(num)
        require('bufferline').buf_exec(num, function(buf, visible_buffers)
            vim.cmd('bdelete '..buf.id)
        end
    end

    vim.cmd [[
        command -count Bdel <Cmd>lua _G.bdel(<count>)<CR>
    ]]
```

The `buf_exec` function no longer exists in the *bufferline* module
The `exec` function in the *bufferline.commands* module seems to be its replacement

### What did you expect to happen?

The documentation should contain:
```vim
CUSTOM-FUNCTIONS                                     *bufferline-functions*

A user can also execute arbitrary functions against a buffer using the
`exec` function in the `bufferline.commands` module. For example

>lua
    require("bufferline.commands").exec(
        4, -- the forth visible buffer from the left
        user_function -- an arbitrary user function which gets passed the buffer
    )

    -- e.g.
    function _G.bdel(num)
        require("bufferline.commands").exec(num, function(bufnr, visible_buffers)
            vim.cmd('bdelete '..bufnr)
        end
    end

    vim.cmd [[
        command -count Bdel <Cmd>lua _G.bdel(<count>)<CR>
    ]]
```
see: [https://github.com/akinsho/bufferline.nvim/blob/ac788fbc493839c1e76daa8d119934b715fdb90e/lua/bufferline/commands.lua#L95](https://github.com/akinsho/bufferline.nvim/blob/ac788fbc493839c1e76daa8d119934b715fdb90e/lua/bufferline/commands.lua#L95)

### Config

N/A

### Additional Information

...

[...]

PR #845, opened 2023-12-12
Update `bufferline-functions` documentation

mcDevnagh:
see #843


akinsho (changes requested):
The `commands` module shouldn't be used by users at the top of the main init file I import all functions that are intended for export. Please use the function exported there and if it's missing please add it back

mcDevnagh:
Missed that re-export.
Should be good to go now

PR #848, opened 2024-01-04
feat(ui): Tab Renaming

muhmud:
* Ported over changes from previous PR

### DIFF
--- a/lua/bufferline/commands.lua
+++ b/lua/bufferline/commands.lua
@@ -9,6 +9,7 @@ local config = lazy.require("bufferline.config") ---@module "bufferline.config"
 local groups = lazy.require("bufferline.groups") ---@module "bufferline.groups"
 local sorters = lazy.require("bufferline.sorters") ---@module "bufferline.sorters"
 local pick = lazy.require("bufferline.pick") ---@module "bufferline.pick"
+local tabpage = lazy.require("bufferline.tabpages") ---@module "bufferline.tabpages"
 
 local M = {}
 
@@ -262,6 +263,17 @@ function M.sort_by(sort_by)
   local opts = config.options
   if opts.persist_buffer_sort then utils.save_positions(state.custom_sort) end
   ui.refresh()
+end
+
+function M.rename_tab(args)
+  if #args == 0 then return end
+  local tabnr = tonumber(args[1])
+  local name = table.concat(args, " ", 2)
+  if tabnr == nil then
+    name = table.concat(args, " ")
+    tabnr = 0
+  end
+  tabpage.rename_tab(tabnr, name)
 end
 
 _G.___bufferline_private.handle_close = handle_close
--- a/lua/bufferline/commands.lua
+++ b/lua/bufferline/commands.lua
@@ -269,7 +269,7 @@ function M.rename_tab(args)
   if #args == 0 then return end
   local tabnr = tonumber(args[1])
   local name = table.concat(args, " ", 2)
-  if tabnr == nil then
+  if not tabnr then
     name = table.concat(args, " ")
     tabnr = 0
   end
--- a/lua/bufferline/tabpages.lua
+++ b/lua/bufferline/tabpages.lua
@@ -42,7 +42,7 @@ function M.rename_tab(tabnr, name)
     name = string(tabnr)
   end
   api.nvim_tabpage_set_var(tabnr, "name", name)
-  M.get()
+  ui.refresh()
 end
 
 function M.get()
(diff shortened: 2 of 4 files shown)


muhmud:
Hey @akinsho , I think this is pretty much ready to look at.

Quick question: how do I run the tests?

lua/bufferline/commands.lua
```diff
@@ -264,6 +265,17 @@ function M.sort_by(sort_by)
   ui.refresh()
 end
 
+function M.rename_tab(args)
+  if #args == 0 then return end
+  local tabnr = tonumber(args[1])
+  local name = table.concat(args, " ", 2)
+  if tabnr == nil then
```
akinsho, line 272:
```suggestion
  if not tabnr then
```

lua/bufferline/tabpages.lua
```diff
@@ -34,6 +34,17 @@ local function render(tabpage, is_active, style, highlights)
...
 end
 
+function M.rename_tab(tabnr, name)
+  if tabnr == 0 then
+    tabnr = vim.fn.tabpagenr()
+  end
+  if name == "" then
+    name = string(tabnr)
+  end
+  vim.api.nvim_tabpage_set_var(tabnr, "name", name)
```
akinsho, line 44:
```suggestion
  api.nvim_tabpage_set_var(tabnr, "name", name)
```

lua/bufferline/tabpages.lua
```diff
@@ -21,7 +21,7 @@ local function render(tabpage, is_active, style, highlights)
   local hl = is_active and h.tab_selected.hl_group or h.tab.hl_group
   local separator_hl = is_active and h.tab_separator_selected.hl_group or h.tab_separator.hl_group
   local chars = constants.sep_chars[style] or constants.sep_chars.thin
-  local name = padding .. tabpage.tabnr .. padding
+  local name = padding .. (tabpage.variables.name or tabpage.tabnr) .. padding
```
akinsho, line 24: I didn't know the `tabpage.variables` existed nice find 👍🏾

lua/bufferline/commands.lua
```diff
@@ -264,6 +265,17 @@ function M.sort_by(sort_by)
   ui.refresh()
 end
 
+function M.rename_tab(args)
```
akinsho, line 268: I think this should maybe have some sort of warning/guard in case someone in buffers mode tries to use this without thinking e.g.

```lua
if config.opts.mode ~= "tabs" then
  vim.notify(... "Don't do this")
end
```

akinsho (commented):
Looks good in general made a few comments. Re. tests I'll run them whenever you push a change since it needs approval to run

lua/bufferline/tabpages.lua
muhmud: Neither did I, the credit for this really goes to @distek

lua/bufferline/commands.lua
muhmud: This still works in buffers mode though, that's how I use it

distek: I'd say having it available in both would be ideal. People who use tab-based workflows as opposed to buffer-based could still rename their tab groups so they have a visual indicator as to what's going on in what tab. And buffer-based workflows could still rename their tab pages.

There might be some logic changes to how tabs are renamed if the mode is set to "tabs", though

lua/bufferline/tabpages.lua
```diff
@@ -34,6 +34,17 @@ local function render(tabpage, is_active, style, highlights)
...
 
+function M.rename_tab(tabnr, name)
+  if tabnr == 0 then
+    tabnr = vim.fn.tabpagenr()
+  end
+  if name == "" then
+    name = string(tabnr)
+  end
+  api.nvim_tabpage_set_var(tabnr, "name", name)
+  M.get()
```
akinsho, line 45: I just noticed this before merging? why does this function need to be called? is this trying to force a refresh? There is a `ui.refresh` function for that if so

akinsho (commented):
Almost there just noticed that the rename function was triggering a get but not sure why?

lua/bufferline/tabpages.lua
distek: I'm assuming my original reasoning was because it was a more tab scoped re-render call? If `ui.refresh` will work fine, though, then that'd make more sense.

muhmud: I'll make this change & test it out

muhmud: Pushed a commit for & tested it out; looks like it's working

akinsho (approved):
LGTM thanks 🚀

PR #852, opened 2024-01-23
fix(tabs): use custom separator_style in tabpages

lucobellic:
Hello,

I would like to use custom `separator_style` table for tabpages so I added the support in this fix.

> [!Note]
> The behavior still differs from buffers since tabpages don't use `ui.add_separators()`, `ui.get_separators()` or other `ui` functions

Thanks


akinsho:
Sorry for the delay. Seems reasonable, thanks

PR #858, opened 2024-01-31
feat(groups): fire event on pin buffer

### LINKED ISSUE #857: [Feature Request]: fire user event on pin buffer
### What?

Dispatch a user event when pinning a buffer. I.e. add something like:
```lua
  vim.api.nvim_exec_autocmds("User", { pattern = "BufferlinePinBuffer" })
```

[to groups.persist_pinned_buffers](https://github.com/akinsho/bufferline.nvim/blob/b15c6daf5a64426c69732b31a951f4e438cb6590/lua/bufferline/groups.lua#L150)

### Why?

I'd like to create a Neotree source that shows pinned buffers, in part to work around #445 

In order to refresh the view, I need to listen on an event. Dispatching a user event from bufferline seems to be the right solution?

jackielii:
fix #857


jackielii:
sorry about the formatting change, it's the stylua auto format on save. If you don't like it, I can revert the formatting changes

Issue #862, opened 2024-02-09
[Feature Request]: Persistent `:BufferLinePick` letters


rish987:
### What?

In buffer-pick mode, currently the letters for each buffer are set on-the-fly, so they will not persist between sessions, meaning that you have to re-learn the letters if there are two filenames that start with the same letter, but you open their buffers in different orders in two separate sessions.

It would be nice if this plugin could remember the letters between sessions. We could invent the concept of "buffer sets" which are associated with the current working directory. Whenever we open a file for the first time in some directory, we assign a letter to it and save this to disk (under this directory's buffer set). If this file has already been assigned in another buffer set, we use the same letter _unless_ it conflicts with one of the letters in the current buffer set, in which case it is re-assigned in all the buffer sets (this is the only case when a letter can be re-assigned) so that there is no conflict.

### Why?

This feature should quickly allow you to pick buffers under the current project without having to look at the label (that is, as soon as you have learned them), and not have to worry about them changing when you quit and re-open vim. We can also think about allowing this to work for buffers that have not yet been opened (in which case bufferline will open them for you).

Issue #315, opened 2022-01-23
feat: Making BufferLinePick use all lowercase choices


jayp0521:
Hello,

I really this plugin and also this feature of using `BufferLinePick`. I was wondering if it would be feasible to make all the choices a user can pick lowercase. Currently, a file named `Buffer.nvim`, would require typing `B` to select that buffer. `b` would be faster in my opinion. I looked through the code, and I think modifying line 219 is all that needs to change, but I never worked with `lua` before, so could be wrong.

https://github.com/akinsho/bufferline.nvim/blob/7b193b1152fcede020d9f69c8239f8a9ef02356f/lua/bufferline.lua#L212-L225

Thanks for reading

PR #863, opened 2024-02-09
feat(pick): persistent pick letters

rish987:
Implements #862

### DIFF
--- a/lua/bufferline/pick.lua
+++ b/lua/bufferline/pick.lua
@@ -14,6 +14,9 @@ local valid = "abcdefghijklmopqrstuvwxyzABCDEFGHIJKLMOPQRSTUVWXYZ"
 
 function M.reset() M.current = {} end
 
+local buffer_sets_file = vim.fn.stdpath("data") .. "/" .. "bufferline-buffer-sets.json"
+M.buffer_sets = vim.fn.filereadable(buffer_sets_file) ~= 0 and vim.fn.json_decode(vim.fn.readfile(buffer_sets_file)) or {}
+
 -- Prompts user to select a buffer then applies a function to the buffer
 ---@param func fun(id: number)
 function M.choose_then(func)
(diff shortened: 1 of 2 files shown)


lua/bufferline/pick.lua
```diff
@@ -32,9 +35,84 @@ function M.choose_then(func)
...
+      set[element.path] = possible_letter
+      M.write_buffer_sets()
+      return possible_letter
+    end
+  end
+
+  -- either buffer never previously assigned or possible_letter already taken in this set, must re-assign (in all buffer sets)
+
+  local letter_order = {}
+  for c in element.name:lower():gmatch("%w") do table.insert(letter_order, c) end
```
rish987, line 93: Oops, sorry I forgot to take out this call to `:lower()`, but perhaps we can talk about that -- I didn't quite understand your concern with #315, if you have `Afile` and `afile`, what is wrong with the label for `afile` being `f`? Is it that it is unexpected? Perhaps with this feature that is less of a problem, since you'll learn that the letter for `afile` is `f` and that will never change.

rish987:
Hi @akinsho, thanks for your comments! To quickly answer some of your initial concerns:

> I don't completely object but also I think the use case seems to me based on my workflow potentially fairly minor? is the main aim to avoid having to press new label for files when you re-open neovim?

Yes, that's all. Well, it's really to decrease the mental overhead of buffer-switching just a little bit more. Once you know a buffer's associated label, you can execute a short mapping to go to it all in one go, and won't have to pause midway to look at the bufferline and verify that you have the right letter (or worse, accidentally press the wrong letter and go to the wrong buffer).

> This seems like quite a large amount of code for a use case that frankly till I saw this PR I wouldn't have imagined necessary and haven't ever had it requested.

Maybe that's partly because `:BufferLinePick` is already a somewhat niche aspect of this plugin. But I think, regardless of whether it has particuarly been brought up before, it seems clear that it is preferrable to not have the buffer-pick letters change between sessions if we can help it.

Anyways, I've realized that this functionality is probably best left to an external buffer-picking plugin, rather than trying to squeeze even more into bufferline. I'm thinking that bufferline could offer some integration with this plugin, that is, querying it to get the letters to display in the bufferline (and falling back to the default scheme for determining the letters if it isn't installed). So, I think I'll close this PR for now and see what I can do to get a plugin of this description working. Let me know if you have any thoughts on the matter -- I will give you an update when I think I have something ready to integrate with bufferline!

PR #873, opened 2024-03-05
fix a typo in rename_tab

archibate:
`string` cannot be called directly:
```
attempt to call a table value (global 'string')
```
To convert a number to string in Lua, we need to call `tostring(num)` instead.


akinsho (approved):
Thanks a bunch for this fix

PR #877, opened 2024-03-07
fix(tabpages): renaming bug on reopened tab

codopanda:
vim.fn.tabpagenr gets the tab number but api.nvim_tabpage_set_var expects tab id, meaning the tab rename would fail if a tab was closed and reopened. This change passes 0 into api.nvim_tabpage_set_var which will use the current focused tab without needing to retrieve the id.

Before this change:

https://github.com/akinsho/bufferline.nvim/assets/159239255/d219e828-6997-4bb0-8fda-07a41c58e085

After this change:

https://github.com/akinsho/bufferline.nvim/assets/159239255/0bf298bd-d6d2-438c-9b5f-c3254397e1b0

PS: Thank you for making bufferline!

### DIFF
--- a/lua/bufferline/tabpages.lua
+++ b/lua/bufferline/tabpages.lua
@@ -35,9 +35,8 @@ local function render(tabpage, is_active, style, highlights)
 end
 
 function M.rename_tab(tabnr, name)
-  if tabnr == 0 then tabnr = vim.fn.tabpagenr() end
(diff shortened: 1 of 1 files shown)


akinsho:
Thanks for the fix @codopanda especially the videos, that makes it super easy to understand what you were solving for.

PR #882, opened 2024-03-13
fix(commands.lua): fix function close_others

minhtri2710:
image
I encountered an error because item.id in state.components is no longer suitable for tabclose to use. I changed the code to be able to close the tabs before and after the current tab. I use Windows OS.

### DIFF
--- a/lua/bufferline/commands.lua
+++ b/lua/bufferline/commands.lua
@@ -248,8 +248,13 @@ function M.close_others()
   local index = M.get_current_element_index(state)
   if not index then return end
 
-  for i, item in ipairs(state.components) do
-    if i ~= index then delete_element(item.id) end
+  local id = 1
+  for i in ipairs(state.components) do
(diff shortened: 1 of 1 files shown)


lua/bufferline/commands.lua
```diff
@@ -248,8 +248,13 @@ function M.close_others()
...
   if not index then return end
 
-  for i, item in ipairs(state.components) do
-    if i ~= index then delete_element(item.id) end
+  local id = 1
+  for i in ipairs(state.components) do
+    if i ~= index then
+      delete_element(id)
+    else
+      id = 2
```
akinsho, line 256: I don't understand what this line of code is trying to do

akinsho (commented):
@minhtri2710 thanks for the PR, can you explain in more detail exactly what you are trying to fix i.e. a step by step of when this issue occurs and why you chose a specific solution so it's clear what is happening

Issue #881, opened 2024-03-08
[Feature Request]: A way to retrieve a buffer's `current_index`


thenbe:
### What?

I wanted to create a keymap to move buffers. Bufferline has a function that allows us to do this:

```lua
require('bufferline').move_to(new_index, current_index)
```

However, I don't know what the buffer's `current_index` is. I know everything about the buffer I want to act on, except it's `current_index`. How can I get a buffer's `current_index` for the purposes of the function above?

Is there some way one can exchange a `bufnr` for a bufferline `current_index`?

### Why?

I'm trying to attach mappings to a telescope buffer picker.

thenbe:
I'm not sure this is a dupe of https://github.com/akinsho/bufferline.nvim/issues/861 because https://github.com/akinsho/bufferline.nvim/issues/861 still assumes we're still *in* the buffer we want to act on AFAICT. Whereas this issue assumes we want to act on "non-current" buffers (i.e. buffers other than the current one).

akinsho:
@thenbe can you please move this discussion either to #861 despite the perceived differences I have too many open issues to keep track of this sort of distinction alternatively please look at #868 to see if that might cover what you want if not maybe comment there

PR #883, opened 2024-03-21
feat: add get_index() to get a buffer's index from its buffer id

### LINKED ISSUE #861: [Feature Request]: Expose buffer index
### What?

expose the get_current_element_index function
add index to get_elements

### Why?

I imagine there are many potential usages, but I'm trying to setup a custom buffer close keybind.
I can submit a pull request if you'd like, but it's an easy enough feature.

EdgarAlenPoe:
Closes #861 and associated request by allowing users to get the index of a given buffer from its buffer id.

This method of solving the issue was taken from #881

### DIFF
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -29,6 +29,7 @@ local M = {
   sort_by = commands.sort_by,
   pick = commands.pick,
   get_elements = commands.get_elements,
+  get_index = commands.get_index,
   close_with_pick = commands.close_with_pick,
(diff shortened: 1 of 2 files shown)


akinsho:
@EdgarAlenPoe thanks for the PR. As I mentioned in the linked issue, I'm not sure I want this API to be  public as I don't think I want users thinking about the internal index of a buffer. I appreciate you have your own specific use case in mind but tbh I'd rather avoid the complexity/issue that comes with exposing this and having to worry about how it will impact future changes I want to make

PR #891, opened 2024-04-21
fix: change from vim.tbl_isarray & vim.tbl_islist to vim.isarray & vi…

dkeithdj:
Neovim build `NVIM v0.10.0-dev-2973+gf694d020c5`
has errors shown below

image

updated the functions like so


akinsho:
Thanks @dkeithdj I've merged #892 it's quite similar but had a version check which is useful to have. Thanks for the PR though

PR #897, opened 2024-04-26
Add annotation to ignore deprecated method

zegabr:
- this is causing errors on nvim-0.10

image

### DIFF
--- a/lua/bufferline/utils/init.lua
+++ b/lua/bufferline/utils/init.lua
@@ -256,6 +256,7 @@ end
 if vim.fn.has("nvim-0.10") == 1 then
   M.is_list = vim.isarray or vim.islist
 else
+---@diagnostic disable-next-line: deprecated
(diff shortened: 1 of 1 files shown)


zegabr:
this is the (nightly) version I'm running
```
nvim -version
NVIM v0.10.0-dev-3007+ga736e845a
Build type: RelWithDebInfo
LuaJIT 2.1.1713484068
```

axelhj:
I stumbled upon this issue today aswell. I found it weird that there is a runtime crash due to a deprecation but I hope this gets merged to fix the issue. I updated to a nightly/master version when I encountered this bufferline crash-on-setup.

akinsho:
I think there is a bit of confusion here. Turning off a diagnostic I don't think will fix this since the issue was actually that a method was deprecated and was still being used. That method is now no longer being used so should not be causing an error. I don't think it's necessary to actually hide this since for me it's a good reminder to come in here and remove this eventually.

PR #900, opened 2024-05-04
doc: clarify bufferline-custom-areas

dmspelma:
Update text in `bufferline-custom-areas` section.

There is an unexplained issue or nuance with the given custom-area example. It is possible for errors to appear with warning/info/hint (or any diagnostic level under any highlight group) due to the way highlight groups are created.

Highlight groups are created on first pass [in L21 of method create_hl](https://github.com/akinsho/bufferline.nvim/blob/main/lua/bufferline/custom_area.lua#L16-L23) It is important to note that they are not unset.

For example, if the first diagnostic is a hint, then highlight group `BufferLineRightCustomAreaText1` will be the hint highlight group. Then, if you fix the hint and introduce an error diagnostic, it will appear with the hint highlighting.

One solution would be to unset highlight groups and then re-set them during method call. Destroying and creating the highlight groups each time tabline is rendered sounds like an expensive solution.

One workaround is to create the table entries with desired highlights so the `BufferLine[side]CustomAreaTest[number]` groups and insert the text if needed.

### DIFF
--- a/doc/bufferline.txt
+++ b/doc/bufferline.txt
@@ -1125,25 +1125,36 @@ to be shown in a list of tables. For example:
         local info = #vim.diagnostic.get(0, {severity = seve.INFO})
         local hint = #vim.diagnostic.get(0, {severity = seve.HINT})
 
+        table.insert(result, {text = "", fg = "#EC5241"})
+        table.insert(result, {text = "", fg = "#EFB839"})
+        table.insert(result, {text = "", fg = "#A3BA5E"})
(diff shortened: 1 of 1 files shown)


gepbird:
I've encountered this issue as well and made a different workaround: https://github.com/akinsho/bufferline.nvim/pull/943/commits/9fedfd15a6bb4683dc894e8d8e4fa98489fea2fa (merged in f98ede02b152a8b86120cda5c806aca3e01e6406).

PR #902, opened 2024-05-14
fix: toggle tabline when tab added w/o new buffer

baodrate:
Currently, when a tab is created without a new buffer being added (e.g. `<C-W>s<C-W>T`), tabline visibility is not toggled. Fix this edge case so that the tabline is displayed when there are either >1 buffers or the tab indicators are enabled and there are >1 tabs.

### DIFF
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -89,7 +89,8 @@ end
 --- If the item count has changed and the next tabline status is different then update it
 local function toggle_bufferline()
   if not config.options.auto_toggle_bufferline then return end
-  local item_count = config:is_tabline() and utils.get_tab_count() or utils.get_buf_count()
+  local item_count = config:is_tabline() and utils.get_tab_count()
+    or utils.get_buf_count() + (config.options.show_tab_indicators and (utils.get_tab_count() - 1) or 0)
   local status = (config.options.always_show_bufferline or item_count > 1) and 2 or 0
   if vim.o.showtabline ~= status then vim.o.showtabline = status end
 end


akinsho:
Tab indicators should be the thing that triggers tabline visibility the tab indicators are about whether or not the indicator for buffer users is visible or not. I don't think I understand this problem entirely since AFAICT, the situation you describe of using this command should still cause the tab count to be > 1 and so that should be enough to trigger visibility

baod-rate:
> the situation you describe of using this command should still cause the tab count to be > 1 and so that should be enough to trigger visibility

Note that the condition is currently `config:is_tabline() and utils.get_tab_count() or utils.get_buf_count()`:

https://github.com/akinsho/bufferline.nvim/blob/0dfc19b7a15a3bc47b975fcffde03859c46dd097/lua/bufferline.lua#L90-L92

When the mode is *not* `tabs`, `toggle_bufferline` *only* considers the number of buffers. So the user opens multiple tabs but with a single buffer, `toggle_bufferline` will never show the tabline, and even though there might be multiple tab indicators visible, the entire tabline is still hidden.

> the tab indicators are about whether or not the indicator for buffer users is visible or not

Yes. I'm checking for this condition because if there are multiple tabs but a single buffer, we only need to display the bufferline if the user enables the tab indicators. This is to avoid the confusing situation of bufferline displaying when there is a single buffer and tab indicators are hidden.

baod-rate:
Okay. My reasoning is that if the user both enabling tab indicators and explicitly opening multiple tabs is a good indication that they'd like to see those tabs. But I respect your judgement. I have this function monkey-patched in my config so I can live with it.

akinsho:
Thanks @baodrate for the contribution, I'll revisit this if a bunch of other people start clamouring for it but in the short term I'd rather avoid changes that could stir up issues or conversation

PR #907, opened 2024-05-18
fix: vim.diagnostic.is_disabled() deprecation warning (deprecated in neovim 0.10.0)

fredrikaverpil:
### Why this change?

Avoid getting deprecation warnings.

### What was changed?

- Replaced `vim.diagnostic.is_disabled()` with `vim.diagnostic.is_enabled()`.
- Removed nil check.
- Removed comment.

### Notes

This requires Neovim 0.10.0: https://neovim.io/doc/user/news-0.10.html

### DIFF
--- a/lua/bufferline/diagnostics.lua
+++ b/lua/bufferline/diagnostics.lua
@@ -71,8 +71,7 @@ local get_diagnostics = {
     local results = {}
     local diagnostics = vim.diagnostic.get()
     for _, d in pairs(diagnostics) do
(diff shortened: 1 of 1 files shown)


folke:
This breaks compatibility with Neovim 0.9.x

akinsho:
I was hoping to avoid having to mediate both versions tbh and rely instead on people who are not ready not move up to stick to a certain version of bufferline where this is still supported. Just back from holiday and catching up on all these breakages so haven't had time to update the docs to point people towards a certain version number to use if you are unable to update to v10

PR #909, opened 2024-05-19
fix: maintain backwards compatibility with neovim 0.9.x

fredrikaverpil:
### Why this change?

As mentioned by @folke in https://github.com/akinsho/bufferline.nvim/pull/907#issuecomment-2119309364, #907 breaks backwards compat with Neovim 0.9.x.

### What was changed?

- Add convenience function to hold the 0.10.0 and 0.9.x syntax for checking whether diagnostics are enabled or not.

### DIFF
--- a/lua/bufferline/diagnostics.lua
+++ b/lua/bufferline/diagnostics.lua
@@ -67,12 +67,21 @@ local function is_insert() -- insert or replace
   return mode == "i" or mode == "ic" or mode == "ix" or mode == "R" or mode == "Rc" or mode == "Rx"
 end
 
+local function diagnostic_is_enabled(d)
+  if vim.fn.has("nvim-0.10") == 1 then
+    return vim.diagnostic.is_enabled({ ns_id = d.namespace, bufnr = d.bufnr })
+  else
+    -- neovim 0.9.x
+    return not (vim.diagnostic.is_disabled and vim.diagnostic.is_disabled(d.bufnr, d.namespace))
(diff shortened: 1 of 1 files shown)


lua/bufferline/diagnostics.lua
```diff
@@ -67,12 +67,21 @@ local function is_insert() -- insert or replace
   return mode == "i" or mode == "ic" or mode == "ix" or mode == "R" or mode == "Rc" or mode == "Rx"
 end
 
+local function diagnostic_is_enabled(d)
+  if vim.fn.has("nvim-0.10") == 1 then
+    return vim.diagnostic.is_enabled({ ns_id = d.namespace, bufnr = d.bufnr })
+  else
+    -- neovim 0.9.x
+    return not vim.diagnostic.is_disabled(d.bufnr, d.namespace)
```
folke, line 75: should be:
```lua
return not (vim.diagnostic.is_disabled and vim.diagnostic.is_disabled(d.bufnr, d.namespace))
```
Older 0.9 versions did not have the method

fredrikaverpil: Thanks, fixed!

akinsho (approved):
Thanks @fredrikaverpil

akinsho:
Thanks both of you 🙏🏾 it's been a pretty eventful return from vacation that it coincided with v0.10.0 and several deprecations 😅

Issue #920, opened 2024-06-07
[Bug]: Plugin broken in neovim 0.10


razziel89:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

The issue is that the plugin no longer works with neovim >= 0.10 because it appears to use a deprecated API call.

### What did you expect to happen?

The plugin was supposed to work with neovim >= 0.10.

### Config

I will skip this section because I will open a PR suggesting a fix soon. I can add it if still required, though.

### Additional Information

[The deprecated API call.](https://neovim.io/doc/user/deprecated.html#vim.tbl_islist())

### commit

5aa803aaa9d8bbbca77d0c8df8ddc0b62c19ff6a

akinsho:
I've responded on the PR itself but I don't think this is still an issue on the most recent version of the plugin. Although tbh I don't think I'm going to try to support nightly anymore anyway. There are way too many changes on there to deal with on a weekly basis

lTriNiTyl:
I think it's also a way to make sure that the files in the .local/share/nvim/site/packer/start/nvim-bufferline.lua/ path are up to date.

akinsho:
@lTriNiTyl that's not how that should work. Your local cached Packer files should not influence how code is written here. You just need to make sure that when you update that properly gets updated and you recompile the Packer file

razziel89:
Hi @akinsho, thanks a lot for your replies, both here and in the PR, and my apologies for causing effort. You are right, the referenced commit is really very old. This appears to be a problem with my package manager, which does not appear to be updating some plugins correctly. I'll close this issue and the PR I've opened.

PR #921, opened 2024-06-07
Fix incompatibility with neovim >= 0.10

razziel89:
Replace deprecated API call by new one. This PR is a suggested fix for #920. I have successfully tested the changed version with neovim >= 0.10 and older versions.

What do you think?

### DIFF
--- a/lua/bufferline/ui.lua
+++ b/lua/bufferline/ui.lua
@@ -116,7 +116,10 @@ end
 ---@param segments bufferline.Segment[]
 ---@return integer
 local function get_component_size(segments)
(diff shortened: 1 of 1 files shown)


akinsho:
@razziel89 this problem has already been fixed. There were several PRs raised around this. If you look at the implementation of `utils.is_list` this has already been done. Are you still actually seeing this error on the latest stable version or in the latest commit? the commit you referenced in your issue is extremely old I think.

razziel89:
Thanks for your replies, @akinsho. You are right, this is not a problem of this plugin but rather with my package manager. Closing this PR.

PR #922, opened 2024-06-09
docs: fix lua codes, consistent nerdfonts symbols, formatting

gepbird:
The last formatting commit 7eb332650221951c5c61adad8f3be6ad2e5a2bcb has a much bigger diff than the other commits, please review commit-by-commit rather than all commits at the same time to make it easier for yourself :)

Let me know if you prefer another style for lua code in docs, I mostly followed the project's `stylua.toml` with a few exceptions:
- 4 space indent (this was more common than 2 spaces)
- don't merge if statements to one line

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,10 +52,10 @@ If you'd like to use an older version of the plugin compatible with nvim-0.6.1 a
 
 ```lua
 -- using packer.nvim
-use {'akinsho/bufferline.nvim', tag = "*", requires = 'nvim-tree/nvim-web-devicons'}
+use { "akinsho/bufferline.nvim", tag = "*", requires = "nvim-tree/nvim-web-devicons" }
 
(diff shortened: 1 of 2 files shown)


akinsho:
Hi @gepbird, 

These appear to be a lot of stylistic changes mixed in with some more important fixes. Can we just address the crucial fixes first in a minimal PR just aimed at those then discuss any style changes as a followup. In general a lot of the attempts to add booleans and things like that are actually contrary to my intentions. The reason I added them like that was actually to specifically discourage copy-pasting which ends up being a huge cause of issues for me since if the code is clearly not copy-pastable people tend to add things in themselves rather than copy a huge blob then complain it doesn't work as they expect

gepbird:
Sure, I'll break these changes up to multiple PRs.

Issue #869, opened 2024-02-27
[Bug]: (UNKNOWN PLUGIN): Error executing lua: attempt to call a number value


pwnalone:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

![image](https://github.com/akinsho/bufferline.nvim/assets/3609458/646f80f5-9b09-4e06-b633-9a8419a0cf72)

### Related Issues

This issue has been reported before three other times that I am aware of, typically in relation to the `nvim-tree` plugin; however, the previous issue authors failed to provide a minimal configuration to help troubleshoot the issue. Furthermore, I have determined that this issue does **not** relate to `nvim-tree` or any other file explorer. Please see below for a minimal configuration.

- #786
- #811
- #820

### Steps to Reproduce

1. Place the minimal configuration at `~/.config/nvim/init.lua`.
2. Start `nvim` to install the 3 relevant plugins.
3. Close and reopen `nvim`.
4. Execute `:h g` and then `:h h` (or anything else that opens up a different help document).
5. Repeat steps 3-4 until you observe the error, above.

##### Dockerfile

Here is a `Dockerfile` to easily setup a clean environment with the latest stable release of Neovim installed.

```dockerfile
FROM ubuntu:latest

ENV DEBIAN_FRONTEND=noninteractive

RUN apt update && apt install -y build-essential curl git tar
RUN curl -sL https://github.com/neovim/neovim/releases/latest/download/nvim-linux64.tar.gz \
    | tar -zxf - -C /usr/local --strip-components=1

CMD /bin/bash
```

##### Build

```bash
docker build -t neovim .
```

##### Start

```bash
docker run --rm -it -v $PWD:/root/.config/nvim -w /root/.config/nvim neovim
```

### What did you expect to happen?

There should not be an error.

### Config

```lua
local lazypath = vim.fn.stdpath('data') .. '/lazy/lazy.nvim'
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({
    'git',
    'clone',
    '--filter=blob:none',
    'https://github.com/folke/lazy.nvim.git',
    '--branch=stable',
    lazypath,
  })
end
vim.opt.rtp:prepend(lazypath)

require('lazy').setup({
  {
    'akinsho/bufferline.nvim',
    version = '*',
    dependencies = { 'nvim-tree/nvim-web-devicons' },
    event = { 'VimEnter' },
    init = function()
      vim.o.termguicolors = true
    end,
[...]

PR #928, opened 2024-06-19
fix: don't reference unloaded buffers

prettymuchbryce:
### Problem
Currently when checking if a buffer is valid, we use the function [nvim_buf_is_valid](https://neovim.io/doc/user/api.html#nvim_buf_is_valid()), however there is an edge case here described in the documentation where a buffer could have been unloaded, but still be "valid". If a buffer is unloaded and valid then referencing `buflisted` will result in a `(UNKNOWN PLUGIN): Error executing lua: attempt to call a number value` error.

### Solution
Instead of `nvim_buf_is_valid` we use [nvim_buf_is_loaded](https://neovim.io/doc/user/api.html#nvim_buf_is_loaded()) which checks both whether a buffer is valid _and_ loaded. This way we never attempt to reference `buflisted` of unloaded or invalid buffers.

This PR should at least partially address #869. Applying the patch removes these errors for my config when invoking Neovim via `nvim .`. Because this error is so generic, it's possible that there could be more than one cause.

### DIFF
--- a/lua/bufferline/utils/init.lua
+++ b/lua/bufferline/utils/init.lua
@@ -145,8 +145,9 @@ M.path_sep = fn.has("win32") == 1 and "\\" or "/"
 --- @param buf_num integer
 function M.is_valid(buf_num)
   if not buf_num or buf_num < 1 then return false end
-  local exists = vim.api.nvim_buf_is_valid(buf_num)
-  return vim.bo[buf_num].buflisted and exists
+  local exists = vim.api.nvim_buf_is_loaded(buf_num)
+  if not exists then return false end
+  return vim.bo[buf_num].buflisted
 end
 
 ---@return integer


akinsho (approved):
Thanks, happy to give this a shot

stevenxxiu:
I just updated my plugins, and this commit causes *possession* to stop working with *bufferline*. With this commit, only the last used tab gets loaded.

latipun7:
> I just updated my plugins, and this commit causes _bufferline_ to stop working with _possession_. With this commit, only the last used tab gets loaded.

I use different session plugin and have same issue.

akinsho:
Sigh 😔 well worth a shot will revert for now, too many obscure requirements to get this working

prettymuchbryce:
> Sigh 😔 well worth a shot will revert for now, too many obscure requirements to get this working 

Rolling back is the right call for now. Clearly this PR causes some unexpected side effects for session plugin users. I am happy to install possession and see if I can further root cause the issue. My initial hunch is that these plugins must intentionally utilize unloaded buffers in a way that I didn't initially anticipate.

akinsho:
Someone suggested swapping the order of the `exists` check as a potential solution which I did. @prettymuchbryce can you see if that made a difference for you?

prettymuchbryce:
> Someone suggested swapping the order of the `exists` check as a potential solution which I did. @prettymuchbryce can you see if that made a difference for you?

Unfortunately that solution does not resolve the `UNKNOWN PLUGIN` errors because that solution still attempts to call `.buflisted` on an unloaded buffer.

I have opened a new PR which I believe contains a working solution. That is— it resolves the `UNKNOWN PLUGIN` errors, while also maintaining support for session plugins that utilize hidden buffers. Let me know your thoughts.

https://github.com/akinsho/bufferline.nvim/pull/931

PR #932, opened 2024-06-25
feat(diag): add `diagnostics_update_on_event` option

### LINKED ISSUE #923: [Bug]: diagnostics can easily get out of date

### What happened?

Bufferline doesn't update when diagnostics change, it can say there are errors when there aren't or vice versa. This is very annoying when using an lsp formatter to fix style errors and seeing that bufferline reports there are still errors (are those actual errors or those errors are outdated? I have to update bufferline to find it out).

With this minimal config, when I delete the erroneous line with `dd`, bufferline still shows the warning until I update it with `:`.

https://github.com/akinsho/bufferline.nvim/assets/29818440/72abd30e-c4e5-4837-89de-8fc0b26faa3a

Bufferline also updates when diagnostics changes.

### Config

```lua
require 'lspconfig'.lua_ls.setup {}
require 'bufferline'.setup {
  options = {
    diagnostics = 'nvim_lsp',
  },
}
```

### Additional Information

Setting `options.diagnostics_update_in_insert = true` for bufferline doesn't solve this issue.

After debugging this issue, I found out that neovim doesn't refresh the tabline when diagnostics change. Bufferline could refresh it, see `:h diagnostic-handlers`. This would make `diagnostic_update_in_insert` deprecated if it was trying to fix this kind of issue. But if `diagnostic_update_in_insert` was introduced to allow reducing tabline redraws, we could introduce an option like `diagnostic_update_on_change`. Please let me know which solution we should go with, I'd be happy to make a PR!

### commit

https://github.com/akinsho/bufferline.nvim/commit/99337f63f0a3c3ab9519f3d1da7618ca4f91cffe

gepbird:
### About this PR

Closes #923.

This change makes the ui refresh whenever nvim reports a diagnostic change.

Demo: bufferline diagnostics are up to date

https://github.com/akinsho/bufferline.nvim/assets/29818440/9a9112ab-c210-4340-8be4-5fd89b46b75d

### Possible improvements for later

When nvim calls our diagnostic handler, we could get extra information about the change, for example the buffer that the change occurred in, all the details about the diagnostics in that buffer and options like `update_in_insert` (configured via `vim.diagnostic.config`). We could use these to replace `vim.diagnostic.get` call (possibly improving performance) and deprecate the bufferline option `diagnostic_update_in_insert`. However it would increase the complexity since we want to keep support for coc, so I decided not to do these changes.

### DIFF
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -181,6 +181,15 @@ local function setup_commands()
   })
 end
 
+local function setup_diagnostic_handler(preferences)
+  if preferences.options.diagnostics == "nvim_lsp" and preferences.options.diagnostics_update_on_event then
+    vim.diagnostic.handlers["bufferline"] = {
+      show = function() ui.refresh() end,
+      hide = function() ui.refresh() end,
+    }
+  end
+end
+
 ---@param conf bufferline.UserConfig?
 function M.setup(conf)
   conf = conf or {}
@@ -192,6 +201,7 @@ function M.setup(conf)
   hover.setup(preferences)
   setup_commands()
   setup_autocommands(preferences)
+  setup_diagnostic_handler(preferences)
   vim.o.tabline = "%!v:lua.nvim_bufferline()"
   toggle_bufferline()
 end
--- a/lua/bufferline/config.lua
+++ b/lua/bufferline/config.lua
@@ -62,6 +62,18 @@ local function validate_user_options(options)
     local item = deprecations[key]
     if item then vim.schedule(function() vim.deprecate(item.name, item.alternative, item.version, "bufferline") end) end
   end
+  if options.diagnostics == "nvim_lsp" and options.diagnostics_update_in_insert then
+    vim.schedule(
+      function()
+        vim.deprecate(
+          "diagnostics_update_in_insert",
+          "vim.diagnostic.config { update_in_insert = true }",
+          "4.6.3",
+          "bufferline"
+        )
+      end
+    )
+  end
 end
 
 ---@param options bufferline.Options
--- a/lua/bufferline/config.lua
+++ b/lua/bufferline/config.lua
@@ -657,6 +657,7 @@ local function get_defaults()
     diagnostics = false,
     diagnostics_indicator = nil,
     diagnostics_update_in_insert = true,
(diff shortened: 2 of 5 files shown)


lua/bufferline/config.lua
```diff
@@ -657,6 +657,7 @@ local function get_defaults()
     diagnostics = false,
     diagnostics_indicator = nil,
     diagnostics_update_in_insert = true,
+    diagnostics_update_on_event = false,
```
akinsho, line 660: I think we can set this to true by default if they are using `nvim_lsp` and then update in insert to false. I'll look into deprecating that eventually

gepbird: Good idea, but shouldn't update in insert be set to what the user configured in nvim with `vim.diagnostic.config`? If so, should we display a warning when a user tries to set update in insert for bufferline config (when using nvim_lsp)?

akinsho: Tbh I'd like to deprecate the setting entirely since it now seems to be moot, if they use update on event then it will update in insert if they have set that in their lsp config. I don't get much time to do granular tweaks often to this plugin so more keen on trying to set the new normal whenever there are changes like this.

Maybe it's worth just add a deprecation warning that shows one time on plugin load to say this option is now essentially irrelevant because you should use update_on_event and make sure that your `vim.diagnostic.config` sets update on insert to true.

I think I have a mechanism in the code base already for deprecating options

gepbird: I like the idea of deprecated that option entirely, but what about coc users?

And yes a deprecation warning on startup would be the best.

gepbird: For now I only deprecated it for nvim_lsp users, so there are no regressions.

lua/bufferline.lua
```diff
@@ -181,6 +181,15 @@ local function setup_commands()
   })
 end
 
+local function setup_diagnostic_handler(preferences)
+  if preferences.options.diagnostics == "nvim_lsp" and preferences.options.diagnostics_update_on_event then
```
akinsho, line 185: nit: this conditional is quite verbose could be local opts = preferences.options

akinsho (approved):

PR #938, opened 2024-07-16
Groups

kuro337:
Mainly added logic to the groups functionality

## Summary of Changes

1. Modified the separator.tab & separator.pill functions to add some features and use the display name if present

2. Cleaned up the render function in groups to remove the redundant loops

3. in the Group Setup - added group specific separator options - such as placing the sep at the start/end to mark visually mark buffers within it

4. Added functionality to Add/Remove groups arbitrarily, so the same tab can go in any group for flexibility

5. Added logic to remove a single tab from a Group

6. Added some helpers for render() and kept the old one for reference/comparison

7. Added a BufferLineDebug user command to print the rendered tabline with Highlights and Text with Padding

8. Updated the doc to provide more info on how to use options [doc/bufferline.txt]

9. Added a  `set_bufferline_hls` function for the user to directly specify all the styles required for Group Labels and Buffers in one go 

10. Fixed a bug with `BufferLineCyclePrev/Next` not working when the current buffer is toggled and in a group

11. Made the type on_close optional in bufferline.Group type

[bufferline-groups.webm](https://github.com/user-attachments/assets/bab5e7d3-34e2-4124-8514-43483e0d96ab)


akinsho:
Hi @kuro337,

Thanks for the PR but I think as it is it is waaaay too big i.e. it introduces way too many changes that need discussion that it would take ages to get through all the points and merge. Can you please do this incrementally. Raise a small PR with the most crucial and minimal of your changes and we can go from there. Realistically based on the time I have available reviewing and commenting on all the changes here will take way too much time for me

doc/tags
akinsho, line 1: This whole file should not be committed I believe this is something generated locally IIRC

lua/bufferline/config.lua
```diff
@@ -671,7 +671,10 @@ local function get_defaults()
     diagnostics_update_in_insert = true,
     diagnostics_update_on_event = true,
     offsets = {},
-    groups = { items = {}, options = { toggle_hidden_on_enter = true } },
+    groups = {
```
akinsho, line 674: Changing defaults like this would require some conversation and convincing.

kuro337:
@akinsho Closing this PR - and created a new PR here!

https://github.com/akinsho/bufferline.nvim/pull/948

PR #941, opened 2024-07-18
feat(tabpages): pass the bufnr to the `name_formatter`

ofseed:
There seems to be no limitation to making `tabs` mode have access to `bufnr`.


akinsho (approved):
LGTM 👍🏿

PR #944, opened 2024-07-18
feat(ui): pass formatter opts to icon fetcher as well

ofseed:
When a user customized the name by `name_formatter`, they would customize the icon too. I think it should be better to make them have the same arguments, but considering changing the arguments would be a breaking change, I recommend adding all the `name_formatter` properties to `get_element_icon`.

In my usage, I use the tab mode of bufferline, I don't want bufferline to show the name of buffers which has a special buftype when there is a normal buffer that exists in the tabpage.

### DIFF
--- a/doc/bufferline.txt
+++ b/doc/bufferline.txt
@@ -136,7 +136,7 @@ The available configuration are:
             },
             color_icons = true | false, -- whether or not to add the filetype icon highlights
             get_element_icon = function(element)
-              -- element consists of {filetype: string, path: string, extension: string, directory: string}
+              -- element consists of {filetype: string, extension: string, directory: string, type:string?}, and all the properties same as the `name_formatter` function
               -- This can be used to change how bufferline fetches the icon
               -- for an element e.g. a buffer or a tab.
               -- e.g.
--- a/lua/bufferline/models.lua
+++ b/lua/bufferline/models.lua
@@ -110,22 +110,23 @@ function Tabpage:new(tab)
   tab.modified = get_modified_state(tab.buffers)
   tab.buftype = vim.bo[tab.buf].buftype
   tab.extension = fn.fnamemodify(tab.path, ":e")
-  tab.icon, tab.icon_highlight = utils.get_icon({
+  ---@type bufferline.TabFormatterOpts
+  local formatter_opts = {
+    name = tab.name,
+    path = tab.path,
+    bufnr = tab.buf,
+    tabnr = tab.id,
+    buffers = tab.buffers,
+  }
+  if tab.name_formatter and type(tab.name_formatter) == "function" then
+    tab.name = tab.name_formatter(formatter_opts) or tab.name
+  end
+  tab.icon, tab.icon_highlight = utils.get_icon(vim.tbl_extend("keep", {
     filetype = vim.bo[tab.buf].filetype,
     directory = fn.isdirectory(tab.path) > 0,
-    path = tab.path,
     extension = tab.extension,
     type = tab.buftype,
-  })
-  if tab.name_formatter and type(tab.name_formatter) == "function" then
-    tab.name = tab.name_formatter({
-      name = tab.name,
-      path = tab.path,
-      bufnr = tab.buf,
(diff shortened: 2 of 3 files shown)


akinsho:
Hi @ofseed,

I'm not sure what you mean? they customize the icon too? as in they should be able to or they have to as well?

ofseed:
When using tab mode, there's `buffers` in `name_formatter`, but there's no equivalent for `get_element_icon`. We can only get information about the currently active buffer

ofseed:
My usage:
```lua
mode = "tabs",
name_formatter = function(args)
  if vim.bo[args.bufnr].buftype ~= "" then
    for _, bufnr in ipairs(args.buffers) do
      if vim.bo[bufnr].buftype == "" then
        return vim.fn.fnamemodify(vim.api.nvim_buf_get_name(bufnr), ":t")
      end
    end
  end
end,
get_element_icon = function(args)
  local loaded, devicons = pcall(require, "nvim-web-devicons")
  if not loaded then
    return ""
  end
  if vim.bo[args.bufnr].buftype ~= "" then
    for _, bufnr in ipairs(args.buffers) do
      if vim.bo[bufnr].buftype == "" then
        return devicons.get_icon(
          vim.fn.fnamemodify(vim.api.nvim_buf_get_name(bufnr), ":t"),
          nil,
          { default = true }
        )
      end
    end
  end
end,
```

ofseed:
> I'm not sure what you mean? they customize the icon too? as in they should be able to or they have to as well?

@akinsho I mean they should be able to, because there's no extra information like buffers opened in the tab in current API, we chan only get current active file's filetype for path, even cannot get it's `bufnr`

doc/bufferline.txt
```diff
@@ -136,7 +136,7 @@ The available configuration are:
             },
             color_icons = true | false, -- whether or not to add the filetype icon highlights
             get_element_icon = function(element)
-              -- element consists of {filetype: string, path: string, extension: string, directory: string}
+              -- element consists of {filetype: string, extension: string, directory: string, type:string?}, and all the properties same as the `name_formatter` function
```
akinsho, line 139: This removes path which is a breaking change and adds type which it doesn't explain here what that is?

lua/bufferline/models.lua
```diff
@@ -166,23 +167,29 @@ function Buffer:new(buf)
...
   if buf.name_formatter and type(buf.name_formatter) == "function" then
-    name = buf.name_formatter({ name = name, path = buf.path, bufnr = buf.id }) or name
+    name = buf.name_formatter(formatter_opts) or name
   end
 
+  buf.icon, buf.icon_highlight = utils.get_icon(vim.tbl_extend("keep", {
+    filetype = vim.bo[buf.id].filetype,
+    directory = is_directory,
+    extension = buf.extension,
+    type = buf.buftype,
```
akinsho, line 190: Why are we passing in the buffer type at all?

doc/bufferline.txt
ofseed: `path` is included in `name_formatter` arguments, which I explained will be fully included. So it simplifies the expression, not changing the behavior.

lua/bufferline/models.lua
ofseed: I kept this because we have done this already, you can see it there:
https://github.com/akinsho/bufferline.nvim/blob/0b2fd861eee7595015b6561dade52fb060be10c4/lua/bufferline/models.lua#L118
Actually, I found it might be a mistake, it may be intended to be used here
https://github.com/akinsho/bufferline.nvim/blob/0b2fd861eee7595015b6561dade52fb060be10c4/lua/bufferline/utils/init.lua#L229
But the `type` here will be the global variable included in Lua stdlib, the `type` function, so it's a false condition. `type` was not documented, but I kept it to avoid underlying breaking change.

akinsho:
@ofseed this will have to wait till I have free time

ofseed:
I just wanted to avoid the message getting lost in the notifications. Thank you.

PR #945, opened 2024-07-18
docs: nerdfonts symbols improvements

gepbird:
Breakdown of #922.

Some nerdfonts symbols got removed, and some just had a different style (eg. filled vs hollow), I decided to go with filled. Github doesn't render nerdfonts symbols, here's how they look: (351763c)
bufferline-nerdfonts-update

cc @akinsho


akinsho (approved):
LGTM 👍🏿

PR #956, opened 2024-08-14
feat: add buffer manager popup window

### LINKED ISSUE #955: [Feature Request]: popup window to reorder/delete buffers
### What?

add support for a popup window that displays all currently open buffers (one filename per line) to reorder buffers and delete buffers. Reordering can be achieved by simply changing the order of the filenames in the popup, deleting by deleting the respective filename/line. The buffers are updated according to the state of the popup window when the popup is closed.

### Why?

When I explore new projects I usually use the LSP to jump from file to file to explore the code. This means that after a while I have a ton of tabs open. Some of them might not be super useful to be open, but are distributed across the bufferline (so I need to cycle through them and close each individual one).

Some other tabs might be far apart from each other, but should ideally be next to each other (e.g. a hpp file and the corresponding cpp file). Moving them over with `BufferLineMove<Prev/Next>` can be quite tedious for a larger amount of buffers.

Inspired by: https://github.com/j-morano/buffer_manager.nvim/tree/main

linusboehm:
Buffers can be re-arranged and deleted in popup. Buffer manager can be opened with `:BufferLineManageBuffers`. Fixes #955

### DIFF
--- a/lua/bufferline.lua
+++ b/lua/bufferline.lua
@@ -10,6 +10,7 @@ local commands = lazy.require("bufferline.commands") ---@module "bufferline.comm
 local tabpages = lazy.require("bufferline.tabpages") ---@module "bufferline.tabpages"
 local highlights = lazy.require("bufferline.highlights") ---@module "bufferline.highlights"
 local hover = lazy.require("bufferline.hover") ---@module "bufferline.hover"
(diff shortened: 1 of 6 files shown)


linusboehm:
@akinsho please let me know, if this is a feature that could potentially be merged. If so, I can update the readme as well.

PR #977, opened 2024-11-11
feat(close): Close current buffer

umkh:
I have used many ways, but they are too inconvenient or do not work properly. I have decided to make this change and added an additional feature.

I added the "**BufferLineCloseCurrent**" command, which I think will be very useful for many people. I would appreciate it if you could review it and share your thoughts with me.

**How it's works?**

https://github.com/user-attachments/assets/2358644e-6ce6-49b3-aa6f-1898c4d356ad

### DIFF
--- a/lua/bufferline/commands.lua
+++ b/lua/bufferline/commands.lua
@@ -276,6 +276,25 @@ function M.rename_tab(args)
   tabpage.rename_tab(tabnr, name)
 end
 
+---Close current buffer 
+function M.close_current()
+  local index = M.get_current_element_index(state)
+  if not index then return end
+  
+  local current_item = state.components[index]
+
+  local length = #state.components
+  if length == 1 then return end
+     
+  local item = state.components[index+1]
+  if not item then 
+    item = state.components[index-1]
+  end
+ 
+  delete_element(current_item.id)
+  open_element(item.id)
+end
+
 _G.___bufferline_private.handle_close = handle_close
 _G.___bufferline_private.handle_click = handle_click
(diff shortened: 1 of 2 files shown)


raymond-van:
Works great for me, thanks.

lua/bufferline/commands.lua
```diff
@@ -276,6 +276,25 @@ function M.rename_tab(args)
...
 
+---Close current buffer 
+function M.close_current()
+  local index = M.get_current_element_index(state)
+  if not index then return end
+  
+  local current_item = state.components[index]
+
+  local length = #state.components
+  if length == 1 then return end
```
ansidev, line 287: @umkh Your proposed command works perfect when there are more than one opening bufferline tabs. However, when only one bufferline tab is opening, this command does nothing. This behaviour is inconsistent with the command `BufferlinePickClose`.
Could you please fix this issue?

ansidev (changes requested):

lua/bufferline/commands.lua
ansidev: I tested these changes. It works.

Replacing

```lua
  if length == 1 then return end

  local item = state.components[index+1]
  if not item then 
    item = state.components[index-1]
  end
```

with

```lua
  local item = current_item
  if length ~= 1 then
    item = state.components[index + 1]
    if not item then item = state.components[index - 1] end
  end
```

umkh: Thank you for your feedback! I appreciate you pointing out the issue. I have already fixed!

PR #998, opened 2025-02-25
feat(duplicated): extract duplicated prefix to constant

iovis:
Extracts the string `(duplicated) ` to a variable in `constants`. This way a user could override it to their preference.

Example:

```lua
require("bufferline.constants").ELLIPSIS = ""
require("bufferline.constants").DUPLICATED = ""
```


akinsho:
@iovis this isn't how people should be changing things in bufferline i.e. exposing internal modules and then changing their values. This is super brittle because I don't expect people to be depending on anything that isn't advertised via the configuration. If I decide to restructure the project and move constants I don't expect to get issues about *my internal project structure* having changed. This should be done via config or not at all ideally

iovis:
would you accept a PR making it configurable?

akinsho:
if the values were passed in via the config table then and worked similar to how the other config options do then sure 👍🏿

iovis:
Sounds like a plan, I'll close this one and work on one where these are configurable, thank you for the feedback!

